Issue #144, opened 2015-03-06
Update transform-common to reflect latest changes in camel-dozer config


kcbabo:
All the details and associated Camel PR can be found here:
https://issues.apache.org/jira/browse/CAMEL-8332

A quick summary of the changes required:
* Add support for expression mappings
* Replace literal support with variable support

kcbabo:
pushed

PR #145, opened 2015-03-10
issue #144 Update transform-common to reflect latest changes in camel-dozer config

kcbabo:
* Switched literal mappings to variable mappings
* Added support for expression mappings


kcbabo:
pushed

PR #175, opened 2015-03-26
Fixes #169: minor issues with new expression support

### LINKED ISSUE #169: minor issues with new expression support
Found a few issues in the fix for 149.  They're all minor, so just lumping them into one GH issue:
* In the UI, expression text is shown surrounded by "{ }".  This is part of the syntax of various languages within Camel (including simple), which will be confusing.  We should not display extra "{ }" to what the user enters.
* Escaping "${" is missing in this commit.  See my comment on the original issue:
https://github.com/fusesource/fuseide/issues/149
* Some expressions could be very long, so viewing the expression text in the source node could be trouble.  Perhaps we should consider having a unique icon/colored box for expressions and just the expression language in there?  We can always provide a mini-viewer of some support in the detail view.
* There's a bug when using an expression in combination with the new "Add a new mapping" support, to reproduce:
    * Click Add a new mapping
    * Set an expression as the source
    * Set a field as the target
    * Boom ...
```
java.lang.ClassCastException: org.jboss.tools.fuse.transformation.editor.internal.util.ExpressionPlaceholder cannot be cast to org.jboss.mapper.Variable
	at org.jboss.tools.fuse.transformation.editor.internal.util.TransformationConfig.update(TransformationConfig.java:497)
	at org.jboss.tools.fuse.transformation.editor.internal.util.TransformationConfig.setTarget(TransformationConfig.java:452)
	at org.jboss.tools.fuse.transformation.editor.internal.MappingViewer.dropOnTarget(MappingViewer.java:91)
	at org.jboss.tools.fuse.transformation.editor.internal.MappingViewer$2.drop(MappingViewer.java:74)
	at org.jboss.tools.fuse.transformation.editor.internal.MappingViewer$DropListener.drop(MappingViewer.java:174)
```

jpav:


kcbabo:
Hey - just a reminder that  you need to delete your existing projects in the IDE and import as existing Eclipse projects after my commit for https://github.com/fusesource/fuseide/issues/159 .  Check the email I sent to the mapper list for more details.

kcbabo:
Reason I point that I out is that there are a bunch of .project files in this PR. :-)

kcbabo:
pushed

jpav:
I have done that already. Is there something that indicates otherwise?

jpav:
Oops, should have kept reading my email :). I'll look at my history, but not sure how that would have happened.

kcbabo:
No biggie.  I didn’t add .project and .classpath files to the other modules outside of the plugin, so you wouldn’t have been able to import those as existing projects anyway.  I will fix that this AM.

jpav:
> On Mar 27, 2015, at 8:10 AM, Keith Babo <notifications@github.com> wrote:
> 
> 
> No biggie. I didn’t add .project and .classpath files to the other modules outside of the plugin, so you wouldn’t have been able to import those as existing projects anyway. I will fix that this AM.

Hmm, I guess I must have done my re-import as existing project efforts after I committed 169.  I know I used the “existing” wizard because I had to go to each folder individually to get the importer to find each nested folder for most of them before I realized there was a “Search for nested projects” button you can click on that wizard.  Oh, well, it was late :-)

Issue #172, opened 2015-03-26
Verify data format IDs exist on "Other" pages of New Transformation wizard


bfitzpat:
There's a strange case now with the revised New Fuse Transformation wizard on the "Other" type for source and target. If you select a Java class and enter anything in the data format ID field, it will let you to the next page or finish. Unfortunately, right now I'm not validating that the ID actually exists we need to either offer a drop-down there or at least validate the text input to make sure that the ID is good.

bfitzpat:
After talking with Keith, we'll make the field (currently a text box) into a drop-down with available data format ids.

bfitzpat:
Created PR https://github.com/fusesource/fuseide/pull/178

bfitzpat:
Can't find a way to decorate the label, but have changed the text color to red if there are no data format definitions available.

lhein:
If you are using data binding you can specify validator and in that validator you could simply return something like return ValidationStatus.warning("bla") to switch to warning icons. It will decorate the txt field automatically. see https://github.com/fusesource/fuseide/blob/master/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/propertysheet/AdvancedEndpointPropertiesSection.java#L296-L316 for example

bfitzpat:
I was trying to avoid using another validator, but that's not a bad way to go.

kcbabo:
pushed

PR #178, opened 2015-03-27
issue #172 - making data format id list a combo and fixing a couple of

bfitzpat:
minor issues with the palette integration


kcbabo:
Lars had previously fixed the NPE issue in the palette extension, so I removed that bit from the PR.

PR #184, opened 2015-03-30
Fixes #182: Set JSON resource type based on selection in new transform wizard

### LINKED ISSUE #182: Set JSON resource type based on selection in new transform wizard
It's too easy to make a mistake on the JSON page of the wizard w/r/t the type of JSON resource definition being used (schema vs. instance document).  The page defaults to schema but the filter needs to match all .json files, so the user can select an instance document by accident (or this is what they intend, but they did not select the instance radio button).  Right now, a mismatch results in really bizarre errors when the wizard completes.  It would be better if we set the radio button state based on the type of JSON file selected.  We used to have code that did this in the old wizard, so we just need to apply the same logic on the JSON page and the radio buttons.

1.  User clicks browse button and selects JSON file.
2.  We look at JSON file content and determine if it's a schema.
3.  If it's a schema, set radio button to schema, else set radio button to instance document.

jpav:


kcbabo:
NewTransformationWizard updates in this PR were creating a nasty merge conflict, so I backed those changes out.  It was just whitespace and removal of SuppressWarnings annotation.

kcbabo:
pushed

Issue #190, opened 2015-03-31
NPE in OtherPage of New Transformation Wizard


kcbabo:
This one is a bit odd.  I was testing the New Transformation Wizard standalone (not through the palette extension) and when I first select it in the list of wizards when I choose File -> New -> Other -> Fuse Transformation and click Next, the following shows up in the Error Log:
```
java.lang.NullPointerException
	at org.jboss.tools.fuse.transformation.editor.internal.wizards.OtherPage.bindControls(OtherPage.java:253)
	at org.jboss.tools.fuse.transformation.editor.internal.wizards.OtherPage.createPage(OtherPage.java:214)
	at org.jboss.tools.fuse.transformation.editor.internal.wizards.OtherPage.access$0(OtherPage.java:127)
	at org.jboss.tools.fuse.transformation.editor.internal.wizards.OtherPage$1.run(OtherPage.java:118)
	at org.eclipse.core.databinding.observable.ObservableTracker.runAndCollect(ObservableTracker.java:150)
	at org.eclipse.core.databinding.ObservablesManager.runAndCollect(ObservablesManager.java:103)
	at org.jboss.tools.fuse.transformation.editor.internal.wizards.OtherPage.createControl(OtherPage.java:114)
	at org.eclipse.jface.wizard.Wizard.createPageControls(Wizard.java:178)
```

kcbabo:
BTW, I can move past this by clicking Next again, but the wizard ends up in a bad state.  Page 2 asks for Java source class when I entered XML as the source type.

bfitzpat:
Issued PR - https://github.com/fusesource/fuseide/pull/194

bfitzpat:
Keith, I have to run but let me know if you run into any issues with these changes

kcbabo:
closed with https://github.com/fusesource/fuseide/pull/196

PR #194, opened 2015-03-31
issue #190 - fixing other page to work outside camel editor

bfitzpat:


kcbabo:
Found a few other issues when testing, so created an updated pull here:
https://github.com/fusesource/fuseide/pull/196

Issue #217, opened 2015-04-14
Enhance Transformation Test Wizard


bfitzpat:
A few things came up in today's meeting:
1) it would be nice if the New Transformation Test wizard actually opened the generated test class
2) It would be good if it handled both Blueprint and Spring Camel route files
3) It would be good if it was more like the Unit Test wizard (either the Eclipse one or the Fuse Test wizard)
4) And as a side note, there was a funky issue that Keith found with the Fuse Test wizard not initializing properly if it was launched from inside an open editor (hot keys for File->New->...) generating an exception and exhibiting odd behavior. Look at that as well.

bfitzpat:
Here's the exception we saw during the meeting today:
'''
!ENTRY org.eclipse.ui 4 0 2015-04-14 13:39:22.546
!MESSAGE Unhandled event loop exception
!STACK 0
java.lang.NullPointerException
	at org.fusesource.ide.branding.wizards.NewCamelTestWizardPageOne.init(NewCamelTestWizardPageOne.java:955)
	at org.fusesource.ide.branding.wizards.NewCamelTestWizard.addPages(NewCamelTestWizard.java:356)
	at org.eclipse.jface.wizard.WizardSelectionPage.getNextPage(WizardSelectionPage.java:116)
	at org.eclipse.ui.internal.dialogs.WorkbenchWizardSelectionPage.getNextPage(WorkbenchWizardSelectionPage.java:101)
	at org.eclipse.jface.wizard.WizardDialog.nextPressed(WizardDialog.java:935)
	at org.eclipse.jface.wizard.WizardDialog.buttonPressed(WizardDialog.java:434)
	at org.eclipse.jface.dialogs.Dialog$2.widgetSelected(Dialog.java:619)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:248)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4353)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1061)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4172)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3761)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:832)
	at org.eclipse.jface.window.Window.open(Window.java:808)
[...]

PR #221, opened 2015-04-15
issue #217 - updated Transformer test wizard and fix for bug found

bfitzpat:
upstream in Camel Fuse code as well


kcbabo:
pushed

Issue #247, opened 2015-04-27
Restrict listing for Camel Context path to just Camel Context definitions


kcbabo:
When the user launches the Fuse Transformation wizard using File -> New, she must select the path to the Camel Context configuration since we don't know which config is being targeted (this is not the case when the wizard is launched from the palette extension).  Right now, we show all XML files, but there is a way (used by the Fuse Transformation Test wizard) to filter the list to only include Camel Context files.

bfitzpat:
Created PR https://github.com/fusesource/fuseide/pull/249 

Not only does this work more like the New Transformation Test wizard (where we check the list of xml files in the workspace or project and only show blueprint or spring-based camel files) but now when the user specifies the camel file and they haven't previously selected a project, we get the project from the selected file and select it for them.

kcbabo:
pushed

PR #249, opened 2015-04-27
issue #247 - changing camel file seletion to be more camel aware, as we

bfitzpat:
did with the test wizard - also changed it so that if the user selects
the camel file and hasn't selected the project, the project is
auto-selected for them


kcbabo:
pushed

PR #306, opened 2015-05-08
issue #305 Pick up @XmlElementDecl types when scanning model classes

### LINKED ISSUE #305: Pick up @XmlElementDecl types when scanning model classes
Depending on the structure of the schema, JAXB will generate classes with @XmlRootElement annotation, ObjectFactory methods with @XmlElementDecl, or both.  Common only detects the former right now, which messes up the UI selection of the target element with the type of schema that produces @XmlElementDecl.

kcbabo:
Fixes #305


kcbabo:
pushed to master and 7.3.x

Issue #314, opened 2015-05-18
Validation logic for collections needs to detect any ancestor, not just parent


kcbabo:
Validation logic is failing in two cases right now:
1) If a source node has a grandparent that is a collection, but a parent that is not a collection, it cannot be mapped to a destination collection.
2) If a source node has a parent that is not a collection, but a grandparent that is a collection, it can be mapped to a target field that does not have a parent collection.

We need to tweak the validation logic to walk back through the ancestry of a node to see if it belongs to a collection.

kcbabo:
pushed to 7.3.x and master

PR #316, opened 2015-05-19
issue #314 Validation logic for collections needs to detect any ancestor, not just parent

kcbabo:


kcbabo:
pushed to 7.3.x and master

PR #322, opened 2015-05-29
FUSETOOLS-1359 - fixing weirdness with relative path in camel file check

bfitzpat:


kcbabo:
pushed to master and 7.3.x

PR #326, opened 2015-06-09
FUSETOOLS-1366 Update transformation common to recognize index mappings

kcbabo:
https://issues.jboss.org/browse/FUSETOOLS-1366


kcbabo:
pushed to 7.3.1 and master

PR #333, opened 2015-07-02
Allows build to at least compile, even if functionality is missing

robstryker:
I've commented out and removed deps on the tm terminal.  This allows your code to actually compile, which means all jars can then be fetched, and all plugins then imported into eclipse for dev work.


lhein:
merged

PR #334, opened 2015-07-02
Refactor the terminal integration

robstryker:
I'd like to make clear, this PR is for ***CONSIDERATION** only.  Do NOT commit it until you test it out. 

Also, the commit message needs to be changed, and I need a JIRA id to link it to. So this patch is a rough draft. 

This patch removes the view plugin entirely. It moves support over to the TM terminal view.


lhein:
merged

PR #335, opened 2015-07-07
FUSETOOLS-1330 - Adding support to select dozer files vs. all xml files

bfitzpat:


bfitzpat:
Merged into master and 7.3.x branches

PR #336, opened 2015-07-07
FUSETOOLS-1333 - Removing "mask" to make errors look like info msgs

bfitzpat:


bfitzpat:
Merged into master and 7.3.x branches

PR #350, opened 2015-08-12
FUSETOOLS-1353 - updates to handle date format

bfitzpat:


transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/MappingDetailViewer.java
```diff
@@ -405,7 +449,7 @@ public void update(final MappingOperation<?, ?> mapping) {
         if (mapping.getType() == MappingType.CUSTOM) createCustomSourcePane(contentPane);
         else {
             final Composite pane;
-            if (mapping.getSource() instanceof Model)
+            if (mapping.getSource() instanceof Model && mapping.getSourceIndex() != null)
```
jpav: I was able to reproduce this after attempting to set the source on an empty mapping.  The real problem isn't here, but in MappingPlaceholder.  It's getSourceIndex and getTargetIndex methods should be checking null.  E.g.:

```
return sourceIndexes == null ? null : new ArrayList<>(sourceIndexes)
```

The change as you have it here will create the wrong diagram for non-indexed fields.

PR #355, opened 2015-08-26
FUSETOOLS-1405 - fixing hidden validation markers in transform wizard

bfitzpat:


bfitzpat:
Merged with 7.3.x and master

PR #363, opened 2015-08-27
FUSETOOLS-1419 - fixing file selection for expression mapping

bfitzpat:


bfitzpat:
Merged with 7.3.x and master

PR #369, opened 2015-09-15
FUSETOOLS-1407 - fixing issue with Other & Java page class browse

bfitzpat:


bfitzpat:
Merged into master and 7.3.x

PR #381, opened 2015-10-01
FUSETOOLS-1463 - brute force workaround for test failure

bfitzpat:


jpav:
Is there a reason to restore this property back to its original value?  I know it’s a new level of “protection”, but we’ve never had this protection before, and anyone that needs to avoid the error will just set the property anyways.  Seems like this would be better done in the constructor without worrying about the restore.

bfitzpat:
I'm good with setting it in the POM, but couldn't figure out a way to do it at our level. As for setting the value back, I like cleaning up my mess.

jpav:
I certainly understand the idea of restoring things back to the way they were before messing with them, but to me this is a mess that Java put in place to begin with.  If we set it in the POM, it would more appropriately be done in the parent POM above us, which is what I had in mind.  If you do keep the current approach of doing it in the code, then we should at least set and unset the property within each method (I thought I remember the PR setting it in generateFromInstance and unsetting it in generateFromSchema).

bfitzpat:
Yeah, I don't think my fix is the best approach by any stretch. If we can do it in the pom, that would be better.

transformation/common/src/main/java/org/jboss/tools/fuse/transformation/model/xml/XmlModelGenerator.java
```diff
@@ -214,6 +216,12 @@ public JCodeModel generateFromInstance(File instanceFile,
                                            File schemaFile,
                                            String packageName,
                                            File targetPath) throws Exception {
+    	// Step 0 - avoid issue with schema loading on Java 8
+    	if (System.getProperty("java.xml.accessExternalSchema") != null) {
```
apupier: is it normal to use java without x here?

transformation/common/src/main/java/org/jboss/tools/fuse/transformation/model/xml/XmlModelGenerator.java
```diff
@@ -214,6 +216,12 @@ public JCodeModel generateFromInstance(File instanceFile,
...
+    		_accessExternalSchemaValue = System.getProperty("java.xml.accessExternalSchema");
+    	}
+        System.setProperty("javax.xml.accessExternalSchema", "all");
```
apupier: Is it this property? https://docs.oracle.com/javase/7/docs/api/javax/xml/XMLConstants.html#ACCESS_EXTERNAL_SCHEMA
Shouldn't we use the constant?

apupier: the constant seems not working

apupier:
I included the modification in https://github.com/fusesource/fuseide/pull/465 because I needed it for fixing the test

PR #397, opened 2015-10-22
FUSETOOLS-1369: Changing attribute server/destination changes the whole SAP endpoint

apodhrad:
This is a workaround before the logic is fully fixed.

### DIFF
--- a/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/propertysheet/PropertiesUtils.java
+++ b/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/propertysheet/PropertiesUtils.java
@@ -311,7 +311,8 @@ public class PropertiesUtils {
     			}
     			if (val.trim().length()<1) val = pparam.getDefaultValue();
     			if (val != null && val.startsWith("/") && !CamelComponentUtils.isFileProperty(pparam)) val = val.substring(1);
-    			if (val != null) syntax = syntax.replace(pparam.getName(), val);
+    			// Workaround for FUSETOOLS-1369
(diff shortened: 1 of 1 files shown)


editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/propertysheet/PropertiesUtils.java
```diff
@@ -311,7 +311,8 @@ public static void updateURIParams(Endpoint selectedEP, UriParameter p, Object v
     			}
     			if (val.trim().length()<1) val = pparam.getDefaultValue();
     			if (val != null && val.startsWith("/") && !CamelComponentUtils.isFileProperty(pparam)) val = val.substring(1);
-    			if (val != null) syntax = syntax.replace(pparam.getName(), val);
+    			// Workaround for FUSETOOLS-1369
+    			if (val != null) syntax = replace(syntax, pparam.getName(), val, syntax.indexOf(':'));
```
lhein: It is not said, that the separator is always the : char. Sometimes there are even multiple different separators used in a single endpoint uri (host:port/path for instance). So your fix will not work in a generic way. I will take a closer look at your commit next week but from first sight I'd say that fix doesn't take care off all the possible edge cases.

apodhrad: I know that ':' is not always the delimeter but in all components listed in current Fuse palette it is used to differ protocol name from the rest of uri.
In your example 'host:port/path' only the part ':port/path' will change which is correct, we don't want to change the protocol name ('host').

lhein:
So it turned out your fix is the solution. I changed it a little bit and pushed it. Thanks for the patch!

PR #413, opened 2016-01-02
FUSETOOLS-1200 Expandable camel-context file in Project Explorer

2point718:
-
Camel context file can now be expanded to see the contained routes in 'Project explorer' tree. Double-click on a route node will load that specific route in the camel editor design page. Route's end-points are shown under the route node to differentiate routes with identical display names.

Notes: I tried to limit changes to the existing code so CamelCtxNavActionProvider 'Open' action lets the editor load the default route and then switches to the selected route. CamelCtxNavRouteNode.getMatchingRouteFromEditorModel performs an approximate equality check on routes by evaluating the display text of nodes. CamelCtxNavContentProvider loads the context file in deferred mode since I wasn't sure about the typical load times of context files.


lhein:
Applied the patch to editor_rework_8.0 branch. This branch will be merged into master in near future.

PR #414, opened 2016-01-08
FUSETOOLS-1233 Open default context file after creation of new project

2point718:
Default camel context file is opened in the editor on completion of the Fuse project wizard. The logic looks for a file named src/main/resources/META-INF/spring/camel-context.xml and if present it will be opened in the editor when the new project wizard exits.

### DIFF
--- a/editor/plugins/org.fusesource.ide.branding/src/org/fusesource/ide/branding/wizards/project/FuseProjectWizard.java
+++ b/editor/plugins/org.fusesource.ide.branding/src/org/fusesource/ide/branding/wizards/project/FuseProjectWizard.java
@@ -19,8 +19,10 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Properties;
 
+import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IProjectDescription;
+import org.eclipse.core.resources.IResource;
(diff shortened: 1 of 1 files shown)


lhein:
Renjith,
thank you for the patch. I have some suggestions here.

1. the master branch is currently not containing the latest code. We do a major refactoring of the model and editor and therefor switched to a topic branch called "editor_rework_8.0". Could you make your contributions using this branch until we merge it back into master? That would help a lot as I don't need to rework your contributions to fit the new model.

2. it is not always the case that the camel context file is in META-INF/spring folder. That only applies to Spring Context files. Blueprint files are stored in OSGI-INF/blueprint. A better approach would be to find all files that fit the camel context content type and then open the first of the found (which is not in the bin or target folder). The logic to find all camel context files of a project should be in the plugin which creates the virtual Camel Contexts folder in the explorer.

Thanks again for all your efforts!

2point718:
Thanks for the support.

1. I will base my work on your topic [branch](https://github.com/lhein/fuseide/tree/editor_rework_8.0). I am closing this PR - will open a new one on the editor_rework_8.0 branch
2. I should be able to look for valid camel files instead of assuming a default path. I looked into that a bit while trying to figure out how to do [FUSETOOLS-1223](https://issues.jboss.org/browse/FUSETOOLS-1223)

PR #418, opened 2016-01-19
JBTIS-438 - add jaxb dependence on jbtis TP instead of local lib

pleacu:


transformation/plugins/org.jboss.tools.fuse.transformation.core/src/org/jboss/tools/fuse/transformation/core/model/json/JsonGenerationConfig.java
```diff
@@ -29,11 +29,6 @@
     private SourceType sourceType = SourceType.JSONSCHEMA;
 
     @Override
-    public Iterator<File> getSource() {
```
lhein: Paul, did u intend to remove this?

transformation/pom.xml
```diff
@@ -131,4 +116,18 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
     </dependencies>
   </dependencyManagement>
 
+  <build>
```
lhein: is that really needed and why?

PR #420, opened 2016-01-19
Put mandatory fields at the top of Properties sections (FUSETOOLS-1586)

apupier:


editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/properties/ParameterPriorityComparator.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ */
+public class ParameterPriorityComparator implements Comparator<Parameter> {
+
+	/* (non-Javadoc)
+	 * @see java.util.Comparator#compare(java.lang.Object, java.lang.Object)
+	 */
+	@Override
+	public int compare(Parameter o1, Parameter o2) {
+		if(isRequired(o1) && isRequired(o2) || !isRequired(o1) && !isRequired(o2)){
```
lhein: shouldn't it be if(  (isRequired(o1) && isRequired(o2)) || (!isRequired(o1) && !isRequired(o2))  ) {

or is that not needed?

PR #427, opened 2016-01-28
Use l10n for error message (see FUSETOOLS-1608)

apupier:


lhein:
Could you please use https://github.com/fusesource/fuseide/blob/master/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/internal/UIMessages.java instead of creating a new class? Thanks!

apupier:
Using UIMessages now :-)

lhein:
good to go

PR #428, opened 2016-01-28
Fusetools 1566 re integrate tests

apupier:


lhein:
Would be good if John and/or Keith could take a look at it.

PR #442, opened 2016-02-17
Support whitespace in workspace path (See FUSETOOLS-1647)

apupier:


apupier:
@lhein Do you have other places in code where this kind of issue can occur?

PR #445, opened 2016-02-18
Use https value for xsd location to avoid error validation by WTP

apupier:
i'm wondering if the documentation is generated? If yes, I shoudl report the fix also on documentation.


lhein:
yep, doc is generated. you should contact Jane.

PR #447, opened 2016-02-22
Select Diagram node when double-clicking on Marker in problem view (see

apupier:
FUSETOOLS-1656)

It redirects to Source editor when there is a line number provided and
no corresponding element found.


lhein:
+1

PR #448, opened 2016-02-22
Avoid NPE when there is no id on a Global Element

apupier:
Should I use <UnknownId> or something else?

### DIFF
--- a/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/CamelContextElement.java
+++ b/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/CamelContextElement.java
@@ -68,6 +68,9 @@ public class CamelContextElement extends CamelModelElement {
 	 */
 	public void addEndpointDefinition(CamelModelElement def) {
 		if (this.endpointDefinitions.containsKey(def.getId())) return;
+		if (def.getId() == null) {
+			def.setId(def.getNewID());
+		}
 		this.endpointDefinitions.put(def.getId(), def);
 		boolean childExists = false;
 		for (int i=0; i<getXmlNode().getChildNodes().getLength(); i++) {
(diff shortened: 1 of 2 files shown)


apupier:
image

lhein:
Every element should have an ID set. We already do that automatically for CamelContext and below but it seems we forgot some elements.

apupier:
In my case I was adding the element in xml directly using the auto-completion.

apupier:
I added more checks. The result is that if there is a missing Id, instead of having NPEs when going to Camel Route Editor or Global Config Editor, the global context element without id are just not referenced in TabContents (which is quite normal as there is no id to reference them)

lhein:
That will force the user to go back into XML and change it manually. Not very user friendly imo. Wouldn't it be easier and better if we just set the id ourselves?

apupier:
It implies doing "magic".
I see 4 options:
- not list them and force user to go back to xml (if he has this issue, he was using the xml before), it is the current PR
- set id ourself when we find missing ones silently, the user will just see a "dirty" editor
- set id ourself when we find missing ones and open a popup to explain user what we are doing (we can even ask him for the ids). it avoids magic stuff.
- Have them listed in Global Config Editor with a validation error and a way to edit them. (my preferred one)

My pov:
- include this PR
- then open a new JIRA enhancement and decide which of the 3 others we want from a functional point of view

What do you think about it? Agree or you prefer that we set silently the id now and open a JIRA enhancement for the two later option?

lhein:
We already do "magic" stuff when the route is opened in the editor. It marks the editor dirty and the user will find out there are now ID fields set. So I'd say we do it similar (means option 2). Though I think it might be nice to have option 4 in this case. So lets raise a new jira for option 4 once u pushed option 2 :)

apupier:
With last provided implementation, the id is automatically provided but the editor is not marked as dirty.

lhein:
+1

apupier:
I'm not very happy with the behavior that I mentioned previsouly:
"With last provided implementation, the id is automatically provided but the editor is not marked as dirty."
I prefer to rethinjk/rework on it before merging it

lhein:
oh, missed that comment. Yeah, we should mark dirty when we change something. Thats maybe something we can do programmatically.

apupier:
this has been covered by https://github.com/fusesource/fuseide/pull/699

PR #450, opened 2016-02-23
Support info and warning validation (see FUSETOOLS-1657)

apupier:
Nota: currently only a single warning validation exists, when ref and uri are specified. i'm not sure about the accuracy of this validation as the uri is set as required. Anyway, the mechanism will be set up


lhein:
+1

PR #451, opened 2016-02-23
URLEncode Path with spaces to retrieve an URI (see FUSETOOLS-1668)

apupier:


apupier:
Other methods calling CamelDebugUtils.getRawCamelContextFilePathFromLaunchConfig seems to not use URi but directly a File so it should be ok for the other ones but I don't know which use cases it is related to exactly in the Fuse IDE so I didn't tested it for other use cases than the one provided on the bug

lhein:
+1

PR #452, opened 2016-02-23
Use provided utility method to ensure having valid launch configuration

apupier:
name (see FUSETOOLS-1647)

It helps to support workspace containing @ and & characters in path.


lhein:
+1

PR #453, opened 2016-02-24
Avoid Missing palette icons (see FUSETOOLS-1569)

apupier:
Just check in the icons folder of the icons. We can't rely on Graphiti
image service as it will provide a "Missing Image as red square" for
everything after the first call to this service.

I think we don't have any icons provided outside from the main plugin right now, right @lhein ?


editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/provider/ImageProvider.java
```diff
@@ -209,10 +206,6 @@ public static String getKeyForLargeIcon(String iconName) {
 	}
 	
 	protected static boolean isImageAvailable(String iconName) {
-		if (CamelUtils.getDiagramEditor() != null) {
-			Image img = GraphitiUi.getImageService().getImageForId(CamelUtils.getDiagramEditor().getDiagramTypeProvider().getProviderId(), iconName);
-			return img != null;
-		}
```
lhein: I know I put that in for a specific reason. Do u simply try to clean up code or does that part cause issues?

apupier: (put here our direct discussion for records)
It is the cause of the issue.
in the code, there was a line searching in graphiti image provider if an image exists for a specific "iconName". the first time it is called, it returns null. The next time it returns a "red square" Image and so it doesn't return anymore null
it was causing the red squares in palette when doing a search

PR #454, opened 2016-02-24
Provide Validation decorator in project explorer view (see

apupier:
FUSETOOLS-1659)

I detected some refresh issue of the markers in Problem explorer view.
Some are related to what I have done and some others were already there (the ones on the xml file)
I can't figure out what is happening. I think it is worthy enough to provide it as is anyway. The user just need to call Refresh (F5) on the project when it occurs.


lhein:
+1

PR #455, opened 2016-02-25
FUSETOOLS-1427 - Refresh Camel Nodes in JMX View on start/stop tracing

apupier:
decorator is now correctly refreshed when using start/stop tracing on Context node AND on Route node


apupier:
@lhein Can you check that there is no other use cases that I missed please?

PS: and sorry to have pushed on the fusesource repo (again) :s

lhein:
looks good to me :D

PR #456, opened 2016-02-26
FUSETOOLS-1574 - Remove all menus which were added before to avoid

apupier:
duplication

Reuse also the Refresh icon provided by the JBoss Tools JMX view

Please note that some selected nodes have no refresh buttons anymore but before the refresh which was available was callign a refresh on another thing so it was working (or not) by side effect only. Will need to open enhancement request for these use cases I think


core/plugins/org.fusesource.ide.foundation.ui/src/org/fusesource/ide/foundation/ui/views/ColumnViewSupport.java
```diff
@@ -301,6 +303,7 @@ protected void addToolBarActions(IToolBarManager manager) {
 
 	protected void removeToolBarActions(IToolBarManager manager) {
 		Menus.removeAction(manager, toolBarActions);
+		Menus.removeAction(manager, refreshAction);
```
lhein: why do u remove it twice here?

apupier: Because I forgot to clean the code correctly after a lot of tested paths :s 
Good catch

PR #458, opened 2016-02-29
FUSETOOLS-1643 - When Nothing is selected on the diagram, hide the

apupier:
Details property section

Introduced an AbstractCamel in order to provide a more modular code. (which introduced most of the file changed)
we also avoid to trigger showing it when Connection and CamelFile are
selected


apupier:
wait for merge: there are failing test when launching in Maven mode but not directly from Eclipse

PR #460, opened 2016-03-01
FUSETOOLS-1562 - Use m2e Maven dependencies

apupier:
using m2e Maven dependencies on a pom provides support for Managed
Dependencies version

/!\ in case the project is not imported in the workspace, if the version comes from a parent pom, the current version will be added (but at least we won't have anymore a NPE and user really should to import its projects)


editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/CamelGlobalConfigEditor.java
```diff
@@ -551,7 +551,7 @@ private void createNewEntry() {
 						List<Dependency> deps = item.getContributor().getElementDependencies();
 						if (deps != null && deps.isEmpty() == false) {
 							try {
-								MavenUtils.updateMavenDependencies(deps);
+								new MavenUtils().updateMavenDependencies(deps);
```
lhein: Just out of curiosity....what is the difference between the 2 calls?

apupier: one of it is unitly testable. Static is the evil for testing :-)

PR #461, opened 2016-03-03
FUSETOOLS-1456 - Search in all folders of the Project for a

apupier:
CamelContentType

It allows to find Camel Context files for Wildfly archetypes too


lhein:
Can you now run the archetype as local camel context? or what is the improvement this pr brings in?

apupier:
The generated project based on the wildfly archetype can be launched via Run as local camel context, but the run will failed due to missing camel dependency.
it allows also to find camel context files that are not in a java src folder.

lhein:
+1

PR #462, opened 2016-03-03
Fusetools 1666 breakpoints update

apupier:
in order to avoid cyclic dependency, I moved the debug part in service core. There are no dependency on UI


lhein:
The breakpoint key used projectName, fileName, endpointNodeId and camelContextId. Do we handle changes to all 4 of these values?

apupier:
We handle endpointNodeId (this PR).
For contextId it seems to work but I suppose that somethign is not very clean.
I created previously a PR for project name: https://issues.jboss.org/browse/FUSETOOLS-1683
and I forgot to create for the file name, here it is: https://issues.jboss.org/browse/FUSETOOLS-1693

Please note that we also don't support element removal when parent container is deleted. I opened another issue too https://issues.jboss.org/browse/FUSETOOLS-1686

apupier:
I think that I will rewrite everything and rely on Eclipse 4 event mechanism (such as in https://github.com/fusesource/fuseide/pull/466) in order to have a clear separation of concerns

apupier:
I use a cleaner way with decoupled concerns in https://github.com/fusesource/fuseide/pull/467

PR #463, opened 2016-03-04
FUSETOOLS-1220 - Register and reuse Launch Configurations for Local

apupier:
Camel Context


lhein:
I find the naming of the launch config a bit confusing:
"Executing clean package in bp for blueprint.xml (Without Tests)"
It would be better to make it something like "Run blueprint.xml as Local CamelContext (without Tests)"

apupier:
updated text

PR #464, opened 2016-03-07
FUSETOOLS-1658 - Avoid overlap for breakpoint/validation decorators and

apupier:
provide a "multi validation decorator"

- Validation decorator are always shifted in order to never have
overlapping
- if several validation issues of a kind are reported, a single
decorator is added with a concatenated tooltip message


lhein:
+1

PR #465, opened 2016-03-07
FUSETOOLS-1616 - Support XSD generation from xml instance containing

apupier:
several namespaces

see also FUSETOOLS-1463

### DIFF
--- a/transformation/plugins/org.jboss.tools.fuse.transformation.core/src/org/jboss/tools/fuse/transformation/core/model/xml/XmlModelGenerator.java
+++ b/transformation/plugins/org.jboss.tools.fuse.transformation.core/src/org/jboss/tools/fuse/transformation/core/model/xml/XmlModelGenerator.java
@@ -23,6 +23,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
 import javax.xml.bind.annotation.XmlElementDecl;
 import javax.xml.bind.annotation.XmlRootElement;
 import javax.xml.namespace.QName;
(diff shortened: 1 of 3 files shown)


lhein:
please ask John / Brian for code review.

transformation/plugins/org.jboss.tools.fuse.transformation.core/src/org/jboss/tools/fuse/transformation/core/model/xml/XmlModelGenerator.java
```diff
@@ -279,7 +287,15 @@ public JCodeModel generateFromInstance(File instanceFile,
         }
 
         // Step 2 - call generateFromSchema with generated schema
-        return generateFromSchema(schemaFile, packageName, targetPath);
+		final JCodeModel generatedSchema = generateFromSchema(schemaFile, packageName, targetPath);
+
+		if (initialValue_AccessExternalSchema != null) {
+			System.setProperty(JAVAX_XML_ACCESS_EXTERNAL_SCHEMA, initialValue_AccessExternalSchema);
+		} else {
+			System.clearProperty(JAVAX_XML_ACCESS_EXTERNAL_SCHEMA);
```
bfitzpat: Considering the discussion we had on FUSETOOLS-1463, do we really need to reset this? Your implementation is definitely a bit cleaner than my hack, so I'm good with it otherwise.

apupier: Personally, I prefer to set it back to its previous value to avoid potential issue with other Eclipse plugins.

bfitzpat: Ok. That was my thinking originally too, so I guess I'm ok with it. :)

+1

PR #468, opened 2016-03-09
FUSETOOLS-1257 - Overrides Launch Configuration with JVM Server Runtime

apupier:
values

### DIFF
--- a/servers/plugins/org.fusesource.ide.server.karaf.core/src/org/fusesource/ide/server/karaf/core/server/subsystems/Karaf2xStartupLaunchConfigurator.java
+++ b/servers/plugins/org.fusesource.ide.server.karaf.core/src/org/fusesource/ide/server/karaf/core/server/subsystems/Karaf2xStartupLaunchConfigurator.java
@@ -12,6 +12,7 @@ package org.fusesource.ide.server.karaf.core.server.subsystems;
 
 import java.io.File;
 import java.io.FileFilter;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


servers/plugins/org.fusesource.ide.server.karaf.core/src/org/fusesource/ide/server/karaf/core/server/subsystems/Karaf2xStartupLaunchConfigurator.java
```diff
@@ -191,10 +182,36 @@ protected String getVMArguments(String karafInstallDir) {
...
+	/**
+	 * @param karafInstallDir
+	 * @param vmLoc
+	 * @return
+	 */
+	String createEndorsedDirValue(String karafInstallDir, File vmLoc) {
+		return String.format("%s%sjre%slib%sendorsed%s%s%slib%sendorsed%s%s%slib%sendorsed", vmLoc.getPath(), SEPARATOR, SEPARATOR, SEPARATOR, File.pathSeparator, vmLoc.getPath(),
+				SEPARATOR, SEPARATOR, File.pathSeparator, karafInstallDir, SEPARATOR, SEPARATOR);
+	}
+
```
robstryker: Eclipse has a "Path" class, which is quite useful  here.  You could try 

new Path(vmLoc.getAbsolutePath()).append("/jre/lib/endorsed/lib/endorsed/lib/endorsed").toOSString()

apupier: ok I will update that on Monday. I just extracted this code and didn't take care of it but effectively it will surely be more readable :-)

lhein:
+1 with Rob's suggestion

robstryker:
The rest of the patch looks great :D

apupier:
I used the java.nio.Paths instead of the Eclipse Path. It consists in the same size of code and relies only on Java and no need to use File separator.

robstryker:
In JBT at least, we try to use eclipse paths and only use raw java / NIO stuff at special locations later on.  For now your change isn't a real issue, but if we ever try to abstract the server adapters out to work for remote servers, a lot of this will need to be changed.

apupier:
In this place, we are providing a path to a VM argument. As the target is outside Eclipse, I would prefer to keep the java.nio.Paths.
@lhein What is your POV?

lhein:
Better fix it right away before we run into issues later.

robstryker:
It's actually not an issue for now and can be fixed later if we get remote working.  It'd be part of a larger fix. There's no real need to fix it now, I just wanted to make you all aware.   We'd need to change to "RemotePath" later anyway.

PR #469, opened 2016-03-10
FUSETOOLS-1691 - Add missing expression kind to the 2.15.2 catalog

apupier:
I added only 3 of them because the last one requires some camel-extra and is deprecated so not sure how to handle it (vtdxml)


lhein:
+1

PR #470, opened 2016-03-10
FUSETOOLS-1474 Provide Only Available Class Dependency In Transformation

apupier:
Java Wizardpage

so now user won't be able to select a "bad class" and then have a CNFE

I think that a bunch of Integration worth to be written, to ensure also no regression on https://issues.jboss.org/browse/FUSETOOLS-1454


apupier:
The trick is to not propose the class that are not in the scope of the project. The CNFE was occuring because we proposed some class that are not visible for the project on which we add the transformation. In fact, the previous was supposedly restricting to the scope of the project BUT it seem sthat there is a bug in JDT and sometimes, it proposes some classes outside of it. I was able to reproduce it when "configure as maven project". Unfortunately I wasn't able to detect the exact root of the issue in JDT itself.
Previously, we were always restricting to Class having java.lang.Object as superclass, no not specifying it just not change anything :-)

bfitzpat:
Interesting. I'll give it a try in the morning and make sure it works for
me locally, but I have to say the changes look pretty straightforward.
Thanks for taking a look at this!

PR #477, opened 2016-03-15
FUSETOOLS-1717 - Support double-click to go to next wizard

apupier:
worth testing on Linux in case events are not triggered in same order


lhein:
+1 - works as expected under Linux

PR #480, opened 2016-03-17
FUSETOOLS-1715 - Provide checkbox to filter on "White list"

apupier:
The white list corresponds to components available in the Palette.

The white list filter is activated by default.

image


lhein:
+1

PR #489, opened 2016-03-30
FUSETOOLS-1533 - Don't override default resources folder

apupier:
after that, it goes further but I have some org.apache.camel.CamelExecutionException: Exception occurred during execution on the exchange: Exchange[message1.xml] with the root cause: "java.lang.ClassNotFoundException: org.apache.camel.spi.DataFormatName"

It should a different issue. what do you think @jpav ?


apupier:
Found the new issue.
It is because the version of Dozer added to the pom is not related to the version of camel.
So the current PR fix the issue.

PR #491, opened 2016-04-08
FUSETOOLS-1767 - Call layout only at the end of the import

apupier:
We are doing a batch import of every elements of the model. Layouting is
needed only at the end of this import.


lhein:
+1
Awesome work!

PR #493, opened 2016-04-13
FUSETOOLS-1789 - Add non-regression test for Unmarshal component

apupier:
- add the test
- extract a Rule to create the project and write more fine grained tests
about Unmarshal component model loading


lhein:
+1

PR #494, opened 2016-04-14
FUSETOOLS-1779: One change is applied more times in SAP components

apodhrad:


core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/util/PropertiesUtils.java
```diff
@@ -470,6 +479,29 @@ public static void updateURIParams(AbstractCamelModelElement selectedEP, Paramet
...
+     * 
+     * @param text text
+     * @param target string to be replaced
+     * @param replacement replacement string
+     * @param delimiter delimiter
+     * @return text with replaced parts  
+     */
+    public static String replaceParts(String text, String target, String replacement, String delimiter) {
+    	String[] parts = text.split(delimiter);
+		for (int i = 0; i < parts.length; i++) {
```
apupier: indentation seems not correct

apupier:
thank you very much for the PR!

We are trying to write tests as far as we can now. I think that it won't be too hard to add unit test in this case.
Can you add them please? I can help you if you need.
(if not possible Lars or I can also work on it)

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/util/PropertiesUtils.java
apodhrad: it should be ok now (I didn't format the whole java code just to keep as small changes as possible)

apupier: seems ok now :-)
it is effectively better to avoid changes everywhere only for formatting reasons.
To apply formatting only on few lines you can select the lines and then use Ctrl+Shift+F

apodhrad:
Unfortunately, I don't know how to easily create an instance of AbstractCamelModelElement :(

apupier:
Effectively this a major flaw for writing unit tests currently.
I think that you should extract 
String newWithoutScheme = replaceParts(withoutScheme, pparam.getName(), val, ":");
// if nothing happens then use the old logic
if (withoutScheme.equals(newWithoutScheme)) {
	withoutScheme = withoutScheme.replace(pparam.getName(), val);
} else {
	withoutScheme = newWithoutScheme;
}

in a dedicated method ( removeScheme() for instance) 
Like that you will get rid of the CamelModelModelElement and be able to write Unit tests for the old way and new way and ensure that both cases are working.

We can also write an integration tests, this time you can have a look to https://github.com/fusesource/fuseide/blob/25e8e18e75d2b67eec5191c17587148028a72bd8/core/tests/org.fusesource.ide.camel.validation.tests.integration/src/org/fusesource/ide/camel/validation/xml/XMLCamelRoutesValidatorIT.java

having both is better but if there are only unit tests for this case it would be enough for me

apodhrad:
I added a very basic test for the logic I changed. I will try to add an integration test. Once I'm done I will let you know. So, please do not push this PR at the moment.

core/tests/org.fusesource.ide.camel.model.service.core.tests/src-test/org/fusesource/ide/camel/model/service/core/util/PropertiesUtilsTest.java
```diff
@@ -0,0 +1,83 @@
+package org.fusesource.ide.camel.model.service.core.util;
```
apupier: don't forget to provide the header also for test files

core/tests/org.fusesource.ide.camel.model.service.core.tests/src-test/org/fusesource/ide/camel/model/service/core/util/PropertiesUtilsTest.java
```diff
@@ -0,0 +1,83 @@
...
+		modelMap = new HashMap<String, String>();
+		modelMap.put("destination", "destination");
+		modelMap.put("idocType", "idocType");
+		modelMap.put("idocTypeExtension", "idocTypeExtension");
+		modelMap.put("applicationRelease", "applicationRelease");
+
+		expected = "sap-idoclist-destination:myDest:idocType:idocTypeExtension:systemRelease:applicationRelease";
+		assertEquals(expected, updatePathParams(syntax, pathParam("destination"), "myDest", pathParams, modelMap));
+
+		expected = "sap-idoclist-destination:destination:abc:idocTypeExtension:systemRelease:applicationRelease";
```
apupier: you should better write separated unit tests for each conditions.
It helps to have a concise view of what is failing exactly
You can @Before method for the initialization or just extract a method
It also helps to understand each tested case as we need to provide a new test method name for each case :-)

apupier:
Thank you for adding the unit tests, you extracted the method at good level i think (better than what I proposed initially ^^)
No problem to wait for the integration test

apodhrad:
during writing the integration test I found another bug in the logic :(

apupier:
That is the nice side of writing tests :-)

apodhrad:
Added also the integration test and fixed the other issue (the problem occured also if there was a slash in the uri)

core/tests/org.fusesource.ide.camel.model.service.core.tests/src-test/org/fusesource/ide/camel/model/service/core/util/PropertiesUtilsTestIT.java
```diff
@@ -0,0 +1,116 @@
+package org.fusesource.ide.camel.model.service.core.util;
```
apupier: don't forget the header on the test file here too

apodhrad:
done

core/tests/org.fusesource.ide.camel.model.service.core.tests/src-test/org/fusesource/ide/camel/model/service/core/util/PropertiesUtilsTest.java
```diff
@@ -0,0 +1,123 @@
+/*******************************************************************************
+ * Copyright (c) 2014 Red Hat, Inc.
```
apupier: we use the creation year, so 2016 instead of 2014

apodhrad:
changed to 2016

apodhrad:
I added one small improvements + one test for using regex. I think that the PR is done. If some changes are needed pls let me know.

core/tests/org.fusesource.ide.camel.model.service.core.tests/src-test/org/fusesource/ide/camel/model/service/core/util/PropertiesUtilsTestIT.java
```diff
@@ -0,0 +1,126 @@
...
+import org.mockito.Mock;
+import org.mockito.MockitoAnnotations;
+
+/**
+ * 
+ * @author apodhrad
+ *
+ */
+@RunWith(Parameterized.class)
+public class PropertiesUtilsTestIT {
```
apupier: The It test need to be placed in another plugin: org.fusesource.ide.camel.model.service.core.tests.integration

apupier:
Thank you for your PR!

PR #500, opened 2016-04-27
FUSETOOLS-1789 - Refactor Property views code

apupier:
- now can write tests on each subparts
- started to factorize code between views for parameters of types
Booleans and Numbers
- it doesn't solve FUSETOOLS-1789 but ease investigation


apupier:
@lhein Can you have a look to that? What do you think about it? It will help to have testable code and factorized code between properties sections. I have done the work for the "easy" part, before going on for the most complicated one, I prefer to have your feedback

lhein:
I like it. Improves readability a lot.

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/properties/creators/AbstractClassBasedParameterPropertyUICreator.java
```diff
@@ -0,0 +1,188 @@
...
+				if (fClass.isInterface()) {
+					wp.setSuperInterfaces(Arrays.asList(fClass.getName()), true);
+				} else {
+					wp.setSuperClass(fClass.getName(), true);
+				}
+				wp.setAddComments(true, true);
+				IPackageFragmentRoot fragroot = null;
+				try {
+					IJavaProject javaProject = (IJavaProject) project.getNature(JavaCore.NATURE_ID);
+					IMavenProjectFacade facade = MavenPlugin.getMavenProjectRegistry().create(project, new NullProgressMonitor());
```
lhein: We might want to change this to use the java classpath instead of the maven one. See https://github.com/fusesource/fuseide/commit/9040b132c0e306c1b462c9d8ea51a010fa41aa85

apupier: right, seems to be related to the FUSETOOLS-1806 issue.
I think we should open another issue for that. Or perhaps a generic one for all search in classpath, maybe with other elsewhere? this one will be just a subtask

apupier: I created https://issues.jboss.org/browse/FUSETOOLS-1818

PR #508, opened 2016-05-04
Use classical Maven structure for unit tests

apupier:
the NumberValidatorTest cannot be launched as Unit test anymore, I moved it to Integration test. i discoverd that some Unit tests were not played due to my bad configuration of the pom.xml, so I change to the classical Maven which finally makes more sense in our use case


lhein:
+1

PR #509, opened 2016-05-04
FUSETOOLS-1824 - extracted xml catalog generator code into own maven module

lhein:


apupier:
Shouldn't we delete the old files? (if I understood well, the code has just been moved in a separate module)

lhein:
we could delete it now or we can just delete it all together later on. That doesn't really matter.

apupier:
if it is possible now, i would prefer. it will help to check that the code has just moved and nothing has changed inside.

lhein:
deleted the old one

PR #510, opened 2016-05-04
Provide a Rule to have access to CamelModel in Unit tests

apupier:
What do you think of having this kind of things?
While writing unit tests I felt the need for it severla times.
i'm not sure that the implementation is the best one.
I required to duplicate the catalog.
For future (or even before the first merge, dependening on discussions):
- ability to define our own catalog to test specific behavior easily
- reuse provided catalog without duplication


core/tests/org.fusesource.ide.camel.model.service.core.tests/META-INF/MANIFEST.MF
```diff
@@ -7,4 +7,5 @@ Fragment-Host: org.fusesource.ide.camel.model.service.core;bundle-version="8.0.0
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Require-Bundle: org.assertj.core;bundle-version="2.1.0",
  org.junit;bundle-version="4.12.0",
- org.jboss.tools.locus.mockito;bundle-version="1.9.5"
+ org.jboss.tools.locus.mockito;bundle-version="1.9.5",
+ org.fusesource.ide.camel.model.service.v2_16_1;bundle-version="8.0.0"
```
lhein: We should try to not use / rely on the service implementation bundles. Service-core should be all we work with.

core/tests/org.fusesource.ide.camel.tests.util/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,10 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: JBoss Fuse Tooling Tests Util
+Bundle-SymbolicName: org.fusesource.ide.camel.tests.util
+Bundle-Version: 8.0.0.qualifier
+Bundle-RequiredExecutionEnvironment: JavaSE-1.8
+Require-Bundle: org.junit;bundle-version="4.12.0",
+ org.fusesource.ide.camel.model.service.core;bundle-version="8.0.0",
+ org.fusesource.ide.camel.model.service.v2_16_1;bundle-version="8.0.0"
```
lhein: We should try to not use / rely on the service implementation bundles. Service-core should be all we work with.

apupier: I don't know how to do otherwise as we can't use OSGi services because we are in Unit test level.

lhein:
I am rather unsure if we can avoid pulling in the 2.16.1 plugin. We should try to though and make it more flexible if possible.

core/tests/org.fusesource.ide.camel.tests.util/META-INF/MANIFEST.MF
lhein: ok, in that case we have no other option I guess

lhein:
+1 for merging

PR #511, opened 2016-05-09
FUSETOOLS-1655 - Include tests into update site

apupier:
- create a feature for core tests
- the goal is to reuse rules in SAP Tooling project
- No unit tests due to limitation of fragment use

### DIFF
--- a/core/features/org.fusesource.ide.core.feature/feature.properties
+++ b/core/features/org.fusesource.ide.core.feature/feature.properties
@@ -23,6 +23,6 @@ providerName=JBoss by Red Hat
 # "description" property - description of the feature
 description=Provides core bundles used by the other features of JBoss Fuse Tooling. This includes Maven support, the documentation, preferences and more.
 copyright=JBoss, Home of Professional Open Source\nCopyright (c) Red Hat, Inc., and individual contributors as indicated\nby the @authors tag, \
-2006-2015. See the copyright.txt in the distribution\nfor a full listing of individual contributors.
+2006-2016. See the copyright.txt in the distribution\nfor a full listing of individual contributors.
 
 licenseURL=license.html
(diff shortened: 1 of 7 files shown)


lhein:
Why do you want to make the tests available in the update site?

apupier:
I have two reasons:
- have acces to some Rules inside SAP Tooling
- start to publish tests as explained here https://github.com/jbosstools/jbosstools-devdoc/blob/master/source/how_to_add_a_test_plugin_or_feature.adoc but in fact seems to be in a separate update site

it seems that I would need more work for this and provide a separate Update Site for the test, need also to check how and where publish it

apupier:
In  https://github.com/jbosstools/jbosstools-devdoc/blob/master/source/how_to_add_a_test_plugin_or_feature.adoc , there are mentions only of the aggregate site of JBoss Tools. there is nothing about the integration stack.
For integrations tack, includign Fuse, it seems that there is https://github.com/jbosstools/jbosstools-integration-stack/blob/master/devstudio/site-ea/category.xml but there is no test yet, will it be the right place? What is the exact link with the update site in our project?
@pleacu maybe you have a more precise idea?

apupier:
after more search, it seems that we should no tuse the integration stack. in fact SAP Tooling is directly linked to Fuse Tooling update site, so we can add th etest on the update site and point SAP tooling to it, which will be http://download.jboss.org/jbosstools/snapshots/builds/Fuse-Tooling/all/repo/ (currently http://download.jboss.org/jbosstools/mars/development/updates/integration-stack/fuse-tooling/8.0.0.Beta2/all/repo )

lhein:
I'd like to hide test features from the users which download/install from the update site. I guess thats also the reason why JBT has a dedicated update site for tests.

pleacu:
Agreed.  Under fuse tooling is the place for it.
               --paull



----- Original Message -----

PR #515, opened 2016-05-13
FUSETOOLS-1805 - created projects now have the configured maven coords

lhein:


apupier:
would be nice to add a test checking for the creation of the project and indeed the value of group and artifact id

lhein:
That whole wizard is going to be abandoned once the new one is ready. Not sure if its worth the time.

apupier:
oh, I thought it was the new one.

PR #516, opened 2016-05-25
FUSETOOLS-1846 - Avoid NPE when no timestamp provided in Exchange

apupier:
- fix Debug mode
- added unit test ensuring creation of Eclipse DebugVariable with emtpy
values
- starting work to have an Integration tests


lhein:
+1

PR #517, opened 2016-05-25
FUSETOOLS-1849 - Expand Tree by default

apupier:
image


lhein:
+1
Could you also remove the "Uncategorized" category from the "Create Global Element" dialog? see FUSETOOLS-1850

PR #518, opened 2016-05-26
FUSETOOLS-1861 - Introduce new Global Definition model element

apupier:
- rely on it for extension points for Global Configuration Editor


lhein:
+1

PR #521, opened 2016-05-31
FUSETOOLS-1827 - Support new Date format from JMX to display in Messages

apupier:
The date format is hardcoded. It must corresponds to the hardcoded value in Camel core https://github.com/apache/camel/blob/camel-2.17.1/camel-core/src/main/java/org/apache/camel/api/management/mbean/BacklogTracerEventMessage.java#L28

What is the best to handle the correlation?


lhein:
Are we still compatible with Fuse Runtime 6.2.0 and 6.2.1 with that code? We need to ensure that.

apupier:
the same format is available since Camel 2.10: https://github.com/apache/camel/blob/camel-2.11.x/camel-core/src/main/java/org/apache/camel/api/management/mbean/BacklogTracerEventMessage.java

lhein:
ok, so should be fine

PR #522, opened 2016-05-31
FUSETOOLS-1862 - Fix hasVariables method

apupier:
add non-regression tests for similar elements


editor/plugins/org.fusesource.ide.launcher/src/org/fusesource/ide/launcher/debug/model/values/CamelExchangeValue.java
```diff
@@ -77,7 +77,7 @@ private void fillfVariables(String variableName, Class<?> type, String value) th
 
 	@Override
 	public boolean hasVariables() throws DebugException {
-		return this.fVariables.isEmpty();
+		return !fVariables.isEmpty();
```
lhein: wow, what a bummer

lhein:
+1

PR #523, opened 2016-05-31
FUSETOOLS-1847 - Reconnect Messages and Diagram View and display all

apupier:
I wasn't able to write Unit tests yet due to JMX Nodes which are connected and that I wasn't able to mock


lhein:
+1

PR #529, opened 2016-06-07
FUSETOOLS-1884 - fixed problem with corrupted cxf config elements

lhein:


lhein:
@apupier could you dbl check this PR?

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
```diff
@@ -658,6 +658,10 @@ protected void setParameter(String name, Object value, boolean overrideChangeChe
 		String javaType = getJavaType(name);
 
 		if (this instanceof CamelContextElement) kind = "attribute";
+
+		// this is needed for FUSETOOLS-1884, otherwise some global config
+		// elements lose their children and get corrupted
+		if (getEipByName(CamelUtils.getTranslatedNodeName(getXmlNode())) == null && name.equalsIgnoreCase("id")) kind="attribute";
```
apupier: Can you explain me why we also need this part?

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/GlobalDefinitionCamelModelElement.java
```diff
@@ -26,4 +26,24 @@ public GlobalDefinitionCamelModelElement(AbstractCamelModelElement parent, Node
 		super(parent, underlyingNode);
 	}
 
+	/* (non-Javadoc)
+	 * @see org.fusesource.ide.camel.model.service.core.model.AbstractCamelModelElement#parseNode()
+	 */
+	@Override
+	protected void parseNode() {
+	}
```
apupier: isn't parseNode enough to override?
Maybe we should add a comment in the empty method to explain why it is empty. something like: "no parsing, reuse XML Node directly"

lhein: I agree

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
lhein: basically because the id parameter is set in any case - even without the parseNode

apupier:
I don't have time to test it on my machine.
it might a manual check that Global COnfig editor is still working for SAP, data formats and Global endpoints.
After that, a little comment for empty method and build finished ok, I'm ok to merge

lhein:
updated PR

lhein:
config elements seem still to work, I will merge the fix after the PR Jenkins is done

PR #532, opened 2016-06-09
FUSETOOLS-1830: If no java source folder exists, the data transformation wizard fails

jpav:
I know there is no test(s) for this, but these changes require PDE tests, and I have to address that in general for the entire project, which will take more time than I want to spend right now for this bug.

### DIFF
--- a/transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
+++ b/transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
@@ -18,9 +18,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URLClassLoader;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
-
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
@@ -30,11 +28,11 @@ import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.IConfigurationElement;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.content.IContentType;
 import org.eclipse.core.runtime.jobs.Job;
-import org.eclipse.jdt.core.IClasspathEntry;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jface.dialogs.MessageDialog;
@@ -166,7 +164,7 @@ public class TransformationEditor extends EditorPart implements ISaveablePart2, 
         // Below is necessary when running from within development Eclipse
         if (buf == null) {
             try (InputStream in =
-                    sourceClass.getResourceAsStream(new Path(Util.RESOURCES_PATH).append(resourcePath).makeAbsolute().toString())) {
+                    sourceClass.getResourceAsStream(new Path(MavenUtils.RESOURCES_PATH).append(resourcePath).makeAbsolute().toString())) {
                 buf = new byte[4096];
                 try (OutputStream out = new FileOutputStream(file)) {
                     for (int len = in.read(buf); len > 0; len = in.read(buf)) {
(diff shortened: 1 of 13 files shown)


transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
```diff
@@ -475,7 +462,7 @@ public void selected(final MappingOperation<?, ?> mapping) {
      */
     @Override
     public void setFocus() {
-    	
+
     }
```
lhein: Do we need to override it?

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/Util.java
```diff
@@ -64,41 +67,35 @@
...
 
-    public static final String JAVA_PATH = MAIN_PATH + "java/"; //$NON-NLS-1$
+    private static final String JAVA_PATH = MAIN_PATH + "java/"; //$NON-NLS-1$
 
     public static final String[] EMPTY_STRING_ARRAY = new String[0];
 
     public static final String TRANSFORMATIONS_FOLDER = ".transformations"; //$NON-NLS-1$
 
-    public static boolean isJSONValid(String incomingJSON) {
+    public static String camelVersion(IProject project) {
```
lhein: please reuse the existing code for this:

https://github.com/fusesource/fuseide/blob/master/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactory.java#L121

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/Util.java
```diff
@@ -681,42 +676,35 @@ private static void updateIndexes(Model model,
 
     public static void updateMavenDependencies(final List<Dependency> dependencies,
```
lhein: If you have a dependency to the camel editor you could also use 

https://github.com/fusesource/fuseide/blob/master/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/utils/MavenUtils.java

We could also move that MavenUtils class to foundation plugin if that helps solving circular dependencies.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/TransformTestWizardPage.java
```diff
@@ -229,7 +229,7 @@ public void widgetSelected(final SelectionEvent event) {
         Dependency dep = new Dependency();
         dep.setGroupId("org.apache.camel"); //$NON-NLS-1$
         dep.setArtifactId("camel-test-blueprint"); //$NON-NLS-1$
-        dep.setVersion(Util.getCamelVersion(project));
+        dep.setVersion(Util.camelVersion(project));
```
lhein: why do u remove the "get" in getter method names? imho using "get" as prefix is the standard way.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/JSONPage.java
```diff
@@ -127,7 +127,7 @@ public void handleValueChange(ValueChangeEvent event) {
                             return;
                         }
                         String jsonText = getJsonText((IFile) resource);
-                        if (!Util.isJSONValid(jsonText)) {
+                        if (!Util.jsonValid(jsonText)) {
```
lhein: why do u remove the "is" in getter method names? imho using "is" as prefix is the standard way for booleans.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/TransformTestWizardPage.java
jpav: Regarding both the getXXX and isXXX comments, of course these are the traditional way of naming methods, but for quite a while I've been annoyed with the useless repetition of the "get" prefix, even though the corresponding "set" prefix is more than enough to distinguish between "read" vs. "write" methods.  The whole paradigm of using the "get" prefix was setup initially to allow introspective code like EE containers to determine which methods are read methods for variable-based properties and not derived methods or those that calculate values, but most people, especially outside of EE code, have since broadened it's use to include all "read"-like methods.  I've started to see a few other API's, including one or two used by the transformation code, that have dropped the use of the "get" prefix, so that discovery, I've been happily following suit.  In fact, you can see that Sun/Oracle itself has followed this approach in the JCodeModel library that we use.  I believe it offers less useless noise to the code and allows the names to more immediately and directly reflect what the method's purpose is.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/Util.java
jpav: I initially wrote my own utilities just to avoid the dependency since much of our code was previously used by a project that wasn't an Eclipse project.  Since it's all Eclipse projects now, I agree I can merge this code.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/Util.java
jpav: Sounds good. I didn't write the initial method; was just moving and renaming it when I ended up modifying it a bit internally.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
jpav: Not sure why that was there.  I'll remove it.

jpav: Oh, never mind, it is required to implement; there just isn't anything really appropriate to give focus when the editor is first displayed.

PR #537, opened 2016-06-12
FUSETOOLS-1665: Remove the new transformation wizard entry from File -> New menu

jpav:


transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/Hl7Page.java
```diff
@@ -117,7 +118,7 @@ public void handleValueChange(ValueChangeEvent event) {
                 }
                 if (path != null) {
                     try {
-                        IResource resource = model.getProject().findMember(path);
+                        IResource resource = CamelUtils.project().findMember(path);
```
apupier: Seems cleaner to be based on the provided model.getProject(). Relying on CamelUtils.Project() introduce some implicit dependency of an Editor already opened and that the one with focus is the correct one.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/JavaPage.java
```diff
@@ -151,14 +152,13 @@ public void widgetSelected(final SelectionEvent event) {
...
                         }
-                        final IType inner = selected;
 
                         UIJob uiJob = new UIJob(Messages.JavaPage_jobName_openError) {
                             @Override
                             public IStatus runInUIThread(IProgressMonitor monitor) {
                                 NewTransformationWizard wizard = (NewTransformationWizard) getWizard();
                                 try {
-                                    Class<?> tempClass = wizard.getLoader().loadClass(inner.getFullyQualifiedName());
+                                    Class<?> tempClass = wizard.loader.loadClass(selected.getFullyQualifiedName());
```
apupier: Any reasons why using the attribute instead of the getter?

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/CamelConfigurationHelper.java
```diff
@@ -10,39 +10,32 @@
...
 public final class CamelConfigurationHelper {
 
-    private CamelFile camelModel;
-    private CamelConfigBuilder configBuilder;
-    private CamelIOHandler handler;
-    
-    private CamelConfigurationHelper(CamelFile camelModel, CamelIOHandler handler) {
-    	this.camelModel = camelModel;
-    	this.handler = handler;
+    public static CamelConfigurationHelper load(File contextFile) {
```
apupier: better to move it below attributes and constructors to follow classical Java conventions

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/Util.java
```diff
@@ -111,14 +109,19 @@ public static boolean draggingFromValidTarget(TransformationManager manager) {
...
         boolean exists = false;
         IClasspathEntry[] entries = javaProject.getRawClasspath();
-        IPath path = javaProject.getPath().append(folder);
+        IPath path = javaProject.getPath().append(folderName);
         for (IClasspathEntry entry : entries) {
             if (entry.getEntryKind() == IClasspathEntry.CPE_SOURCE && entry.getPath().equals(path)) {
                 exists = true;
+                File folder = javaProject.getResource().getLocation().append(folderName).toFile();
+                if (!folder.exists()) {
+                    if (!folder.mkdirs()) throw new Exception(Messages.bind(Messages.Util_UnableToCreateSourceFolder, folder));
```
apupier: It might be best to use an IOException instead of the generic Exception
We could throws IOException and CoreException on the method.

apupier: better to go to a new line and use { }

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/Hl7Page.java
jpav: That was the entire point of this issue, to make the transformation editor only accessible via the Camel editor.  In fact, we had a couple of other issues logged that arose because we, some time after the new editor framework was introduced, changed code that assumed that already assumed an open and focused Camel editor.  Thus, things broke when the transformation editor was accessed directly via the new Fuse transformation action.  I was able to close those out as duplicates by removing that action in this fix and going with the idea that the Camel editor must be the only thing that launches the transformation editor.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/JavaPage.java
jpav: Because previously, when the project was unknown or modifiable, there used to be a getter that had more than a single line of logic.  But now I do notice that there are several variables in that wizard marked as public that should be immutable after the wizard is initialized.  I'll fix that.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/CamelConfigurationHelper.java
jpav: Not sure what you mean by that.  Static members should always be above instance members to highlight their independence from object construction.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/Hl7Page.java
apupier: ok

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/CamelConfigurationHelper.java
apupier: I used to avoid to mix static and non-static so not sure.
SonarLint reporting it as a (little) issue but it might be a bad convention that they are pushing.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/Util.java
jpav: Regarding the exception, unless we're going to handle the exception differently, I see no point to being specific about what we use.  There are already more than one exception thrown by the method, all of which indicate a fatal situation that gets logged, displayed, and prevents to user from proceeding (In the editor, they all get generically wrapped in a PartInitException anyways).  If we ever have some need to make a delineation, I can change this.

Regarding the new line, that's really a style, and therefore, religious :-), view. I prefer single-line statements when possible for readability, code brevity, and the slight increased editing productivity.

apupier: IOException is more explicit than a common Exception.
Having specific exception thrown will help us to wrap them and provide specific feedback later on. using generic Exception we will have to dig through the whole code which exceptions can really be thrown.
I would prefer to do it from the start (but I won't stop the merge of the PR for that)

For multiple statements on a single line, using them is religious but having them on a separated line is scientific. It greatly improves readability and avoids ambiguity. All software quality tools that I met enforced this way of doing it. (again i won't block the PR for this point)

jpav: I would normally agree regarding a more specific exception, but in this case we already know up front that if the source folder cannot be created or determined for _any_ reason, the wizard cannot continue, regardless of whether the cause is due to a system I/O problem or something else.  The only important distinction will ever be the error message.  IOW, I'm confident there will be no later on in this case.

Regarding the style issue, I'm not sure how the traditional approach is any more scientific.  I'm arguing that the single line is actually more readable and avoids ambiguity.  I believe the single line makes the statement visually flow better as read and makes it very clear the if statement has no additional content (else it would appear in a block).  There's a reason the language allows for statements to be made outside of blocks.

PR #538, opened 2016-06-13
removing the unused project wizard (FUSETOOLS-1868)

lhein:


apupier:
+1 when build has finished

PR #539, opened 2016-06-13
extracted weight and name into interface and created comparator (FUSE…

lhein:
…TOOLS-1871)

### DIFF
--- a/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/model/NameAndWeightComparator.java
+++ b/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/model/NameAndWeightComparator.java
@@ -1,0 +1,32 @@
+/******************************************************************************* 
+ * Copyright (c) 2016 Red Hat, Inc. 
+ * Distributed under license by Red Hat, Inc. All rights reserved. 
+ * This program is made available under the terms of the 
+ * Eclipse Public License v1.0 which accompanies this distribution, 
+ * and is available at http://www.eclipse.org/legal/epl-v10.html 
+ * 
+ * Contributors: 
+ * Red Hat, Inc. - initial API and implementation 
+ ******************************************************************************/ 
+package org.fusesource.ide.projecttemplates.wizards.pages.model;
+
+import java.util.Comparator;
+
+/**
+ * @author lhein
+ */
+public class NameAndWeightComparator implements Comparator<NameAndWeightSupport> {
+
+	/* 
+	 * (non-Javadoc)
+	 * @see java.util.Comparator#compare(java.lang.Object, java.lang.Object)
+	 */
+	@Override
+	public int compare(NameAndWeightSupport o1, NameAndWeightSupport o2) {
+		int res = Integer.compare(o1.getWeight(), o2.getWeight());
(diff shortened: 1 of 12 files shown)


editor/tests/org.fusesource.ide.projecttemplates.tests/src/test/java/org/fusesource/ide/projecttemplates/NameAndWeightComparatorTest.java
```diff
@@ -0,0 +1,65 @@
...
+	 * tests the case where the weights are equal but names differ
+	 * 
+	 * @throws Exception
+	 */
+	@Test
+	public void testEqualWeightDifferentName() throws Exception {
+		NameAndWeightSupport o1 = new CategoryItem("o1", "test1", 10, null);
+		NameAndWeightSupport o2 = new CategoryItem("o2", "test2", 10, null);
+		int res = comp.compare(o1, o2);
+		assertEquals(-1, res);
```
apupier: I see that you added AssertJ in dependencies so you can leverage it using assertThat(res).isEqualTo(-1)
at the beginning I was not very convinced about this syntax but after using it several weeks, I find it really nicer, even more when you work with arrays or specific types (Date). For integer the interest is limited to keeping consistency with framework language in other project.
So I propose either to use ASsertJ assertions or to remove the dependency in the plugin.xml

editor/tests/org.fusesource.ide.projecttemplates.tests/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,11 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: JBoss Fuse Tooling Project Templates Tests
+Bundle-SymbolicName: org.fusesource.ide.projecttemplates.tests
+Bundle-Version: 8.0.0.qualifier
+Fragment-Host: org.fusesource.ide.projecttemplates;bundle-version="8.0.0"
+Bundle-RequiredExecutionEnvironment: JavaSE-1.8
+Require-Bundle: org.junit;bundle-version="4.12.0",
+ org.assertj.core;bundle-version="2.1.0",
+ org.jboss.tools.locus.mockito;bundle-version="1.9.5",
```
apupier: I think no Mockito tests is curently used in this plugin, we might remove it

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/model/NameAndWeightComparator.java
```diff
@@ -0,0 +1,30 @@
...
+public class NameAndWeightComparator implements Comparator<NameAndWeightSupport> {
+
+	/* 
+	 * (non-Javadoc)
+	 * @see java.util.Comparator#compare(java.lang.Object, java.lang.Object)
+	 */
+	@Override
+	public int compare(NameAndWeightSupport o1, NameAndWeightSupport o2) {
+		int res = Integer.compare(o1.getWeight(), o2.getWeight());
+		if (res == 0) res = o1.getName().compareTo(o2.getName());
```
apupier: don't you prefer to use a single statement per line?

apupier:
+1 when build finished

PR #544, opened 2016-06-15
FUSETOOLS-1895 - fixed doubled entries in the plugin.xml of projects …

lhein:
…bundle

### DIFF
--- a/editor/plugins/org.fusesource.ide.project/plugin.xml
+++ b/editor/plugins/org.fusesource.ide.project/plugin.xml
@@ -87,28 +87,6 @@
 	</extension>
 
  <extension
-       point="org.eclipse.ui.commands">
-    <command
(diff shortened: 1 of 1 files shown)


editor/plugins/org.fusesource.ide.project/plugin.xml
```diff
@@ -87,28 +87,6 @@
 	</extension>
 
  <extension
-       point="org.eclipse.ui.commands">
```
apupier: I don't see another extension point for this command and handler. As the commit mentioned a "fixed double entries". I just want to check that you want to delete them. (defined in another plugin perhaps?)

lhein: I already deleted the other ones and thought thats it but it seems many entries here have been duplicated which then caused the issue.

PR #546, opened 2016-06-16
FUSETOOLS-1893 - Use Model Name of data format

apupier:
- the model name corresponds to the EIP pattern
- set default value as attributes to handle "specialized" dataformats
- improve DataFormatContributor.canHadle to avoid dataformats not
recognized and put in "Uncategorized"


lhein:
I am not sure I like the manually changed catalog.xml files. It will cause us to do those changes for every new release of Camel we want to support and I fear we might miss that.

apupier:
me neither, i created a task for CAMEL: https://issues.apache.org/jira/browse/CAMEL-10062
I asked for help on IRC but i had no answer

lhein:
+1

PR #547, opened 2016-06-17
FUSETOOLS-1901 - Remove some logs

apupier:
these logs occurs very often and have little interest

@lhein unless I miss a point?


apupier:
I updated to log only in debug mode

lhein:
+1

PR #551, opened 2016-06-22
FUSETOOLS-1872 - this event handler is not used anymore

apupier:


lhein:
I don't like the idea to remove that. We will most probably need that in future so why not just keeping it but in a state where its doing nothing.

apupier:
if we need it, we can add it back and it is not lost, it is available in git history. Also, this extension point is deprecated and we should use the new provided extension point org.eclipse.wst.common.project.facet.core.listeners

lhein:
+1

PR #555, opened 2016-06-27
FUSETOOLS-1922 - Avoid to use File from bin folder

apupier:
there was already target folder ignored, need to ignore also bin one

it will avoid the random failures on CI (and fix a real issue)


lhein:
+1

PR #557, opened 2016-06-27
FUSETOOLS-1920 - Avoid NPE when there is no model linked to Editor

apupier:
I suppose that this use case might happen when there is an invalid xml
file.


apupier:
same as https://github.com/fusesource/fuseide/pull/553 but on the correct repo/branch...

lhein:
:) +1

PR #558, opened 2016-06-28
FUSETOOLS-1918 - Update samples to be deployed on Fuse

apupier:
- configure Maven bundle plugin to generate a valid MANIFEST.MF in all
provided samples
- Configure Fuse runtime to be able to launch fuse.camel type (even if
not working yet)
- improve robustness to bad configurations in several places

Even if not fully functional, at least goes to a better state.
Now be able to deploy jst.utility Module of all blueprint samples to Fuse Runtime even if there are spaces


servers/plugins/org.fusesource.ide.server.karaf.core/src/org/fusesource/ide/server/karaf/core/publish/jmx/OSGIMBeanPublishBehaviour.java
```diff
@@ -47,17 +47,20 @@ public long getBundleId(MBeanServerConnection mbsc,
...
-						}	
-					} else {
-						// if we don't have a version we take the first best
-						if (bsn.equals(bundleSymbolicName)) {
-							return Long.parseLong(id);
+					Object symbolicName = cd.get("SymbolicName");
+					if(symbolicName != null){
+						String bsn = symbolicName.toString();
+						String id = cd.get("Identifier").toString();
+						String ver = cd.get("Version").toString();
```
robstryker: You sure this doesn't need npe check like symbolic link?

robstryker: You sure this doesn't need npe check like symbolic link?

servers/plugins/org.fusesource.ide.server.karaf.core/src/org/fusesource/ide/server/karaf/core/publish/jmx/OSGIMBeanPublishBehaviour.java
```diff
@@ -47,17 +47,20 @@ public long getBundleId(MBeanServerConnection mbsc,
...
-							return Long.parseLong(id);
-						}	
-					} else {
-						// if we don't have a version we take the first best
-						if (bsn.equals(bundleSymbolicName)) {
-							return Long.parseLong(id);
+					Object symbolicName = cd.get("SymbolicName");
+					if(symbolicName != null){
+						String bsn = symbolicName.toString();
+						String id = cd.get("Identifier").toString();
```
robstryker: You sure this doesn't need npe check like symbolic link?

apupier: I didn't encounter the case during development but it can be proper. I will check for it too

servers/plugins/org.fusesource.ide.server.karaf.core/src/org/fusesource/ide/server/karaf/core/server/subsystems/Karaf2xPublishController.java
```diff
@@ -213,12 +209,20 @@ public int transferBuiltModule(IModule[] module, IPath srcFile, IProgressMonitor
 
 	@Override
 	public int removeModule(IModule[] module, IProgressMonitor monitor) throws CoreException {
-		BundleDetails bd = new ModuleBundleVersionUtility().getBundleDetails(module, null);  // This will clearly break... wtf.  Crap. 
+		IPath tmpArchive = getTempBundlePath(module);
+		
+		BundleDetails bd = new ModuleBundleVersionUtility().getBundleDetails(module, tmpArchive);  // This will clearly break... wtf.  Crap.
```
robstryker: If the temporary archive doesn't exist, and it's an osgi bundle rather than a fuse, it will fail to read the information from the osgi file and this will fail.

apupier: So need to implement it. i think that pure osgi file undeploy was not working either. Need to test it.

apupier: I suppose that it is already supported by the ModuleBundleVersionutility. It depends on the Module type deployed if we look at the implentation here: https://github.com/fusesource/fuseide/blob/master/servers/plugins/org.fusesource.ide.server.karaf.core/src/org/fusesource/ide/server/karaf/core/server/subsystems/publish/ModuleBundleVersionUtility.java#L45

apupier: it was working for pure OSGi when the version has no .qualifier. I implemented it to have it working with .qualifier too (by just not checking version in this case)

lhein:
+1

apupier:
we still have an issue when the project has never been "Maven built" and there is no files in target folder but I created a separated issue: https://issues.jboss.org/browse/FUSETOOLS-1937

PR #560, opened 2016-07-01
FUSETOOLS-1856: Allow to specify what kind of information will be displayed

apodhrad:


core/tests/org.fusesource.ide.camel.model.service.core.tests.integration/src/main/java/org/fusesource/ide/camel/model/service/core/tests/integration/core/model/CamelBasicModelElementTestIT.java
```diff
@@ -0,0 +1,66 @@
+package org.fusesource.ide.camel.model.service.core.tests.integration.core.model;
```
apupier: missing header

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/UserLabelsListEditor.java
```diff
@@ -0,0 +1,123 @@
+package org.fusesource.ide.camel.editor.preferences;
```
apupier: missing header

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/UserLabelsListEditor.java
```diff
@@ -0,0 +1,123 @@
...
+ */
+public class UserLabelsListEditor extends ListEditor {
+
+	public static final IInputValidator VALIDATOR = new IInputValidator() {
+		@Override
+		public String isValid(String newText) {
+			if (newText.matches(USER_LABEL_REGEX)) {
+				return null;
+			}
+			return "Input is not valid!";
```
apupier: Externalize string

editor/tests/org.fusesource.ide.camel.editor.tests.integration/src/main/java/org/fusesource/ide/camel/editor/integration/preferences/UserLabelsListEditorTestIT.java
```diff
@@ -0,0 +1,84 @@
+package org.fusesource.ide.camel.editor.integration.preferences;
```
apupier: missing header

editor/tests/org.fusesource.ide.camel.editor.tests.integration/src/main/java/org/fusesource/ide/camel/editor/integration/preferences/UserLabelsListEditorTestIT.java
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	@Before
+	public void createListEditor() {
+		listEditor = new UserLabelsListEditor(PreferencesConstants.EDITOR_USER_LABELS, "", shell);
+		listEditor.setPreferenceStore(PreferenceManager.getInstance().getUnderlyingStorage());
+	}
+
+	@Test
+	public void testValidInputs() {
```
apupier: I think that this test can be moved to Unit test instead of integration test (faster feedback)

editor/tests/org.fusesource.ide.camel.editor.tests.integration/src/main/java/org/fusesource/ide/camel/editor/integration/preferences/UserLabelsListEditorTestIT.java
```diff
@@ -0,0 +1,84 @@
...
+		Assert.assertNull(UserLabelsListEditor.VALIDATOR.isValid("abc.xyz"));
+		Assert.assertNull(UserLabelsListEditor.VALIDATOR.isValid("123.321"));
+		Assert.assertNull(UserLabelsListEditor.VALIDATOR.isValid("abc-2.xyz"));
+	}
+
+	@Test
+	public void testInvalidInputs() {
```
apupier: I think that this test can be moved to Unit test instead of integration test (faster feedback)

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/UserLabelsListEditor.java
```diff
@@ -0,0 +1,123 @@
...
+		getUpButton().dispose();
+		getDownButton().dispose();
+		editButton = createEditButton(buttonBox, "ListEditor.edit");//$NON-NLS-1$
+		editButton.moveAbove(getRemoveButton());
+		return buttonBox;
+	}
+
+	private Button createEditButton(Composite parent, String key) {
+		Button button = new Button(parent, SWT.PUSH);
+		button.setText("Edit");
```
apupier: Externalize String

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/UserLabelsListEditor.java
```diff
@@ -0,0 +1,123 @@
...
+	public Composite getButtonBoxControl(Composite parent) {
+		Composite buttonBox = super.getButtonBoxControl(parent);
+		getUpButton().dispose();
+		getDownButton().dispose();
+		editButton = createEditButton(buttonBox, "ListEditor.edit");//$NON-NLS-1$
+		editButton.moveAbove(getRemoveButton());
+		return buttonBox;
+	}
+
+	private Button createEditButton(Composite parent, String key) {
```
apupier: the key parameter is not used, right?

core/tests/org.fusesource.ide.camel.model.service.core.tests.integration/src/main/java/org/fusesource/ide/camel/model/service/core/tests/integration/core/model/CamelBasicModelElementTestIT.java
```diff
@@ -0,0 +1,66 @@
...
+	@Before
+	@After
+	public void cleanUserLabelsPreference() {
+		PreferenceManager.getInstance().savePreference(EDITOR_USER_LABELS, "");
+		PreferenceManager.getInstance().savePreference(EDITOR_PREFER_ID_AS_LABEL, false);
+	}
+
+	@Test
+	public void testDisplayingDefaultText() {
+		Assert.assertEquals("Log coolName", simpleLog().getDisplayText());
```
apupier: you can use static import for Assert.assertEquals (or even Assertions.assertThat if using AssertJ API)
See http://tools.jboss.org/blog/working-efficiciently-junit-eclipse-update.html how to configure automatic static import
(or Ctrl+Shift+M if I remember well when it was already written)

apodhrad:
All suggestions were applied

apodhrad:
Also suggestion about using static methods was applied

apupier:
Functional feedback:
- in preference page, we need to provide more explanation on "preferred Labels".
-  the "If enabled the ID values will be used for labels if existing" should be closer to "Preferred Labels"
image
- we may also disable the Preferred Labels when  "If enabled the ID values will be used for labels if existing" is checked
- When opening the "New..." input dialog, we have "Input is not valid!"
-- better to provide a default value to avoid it
-- provide a bette rmessage: "Cannot be empty", "The input must have two parts separated by a '.'" or something like that
-- the message should better explained how to what it refers <componentScheme>.<attributeName> (is it right?)
- Would be nice to provide a sort on the Preferred Labels (or the order has an importance? in this case we should provide move up and down button)
- I was able to have it working only to "log" component, is it because it has the scheme declared directly and not a to and only the scheme in the uri attributes?

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/UserLabelsListEditor.java
```diff
@@ -0,0 +1,134 @@
...
+
+	private Button editButton;
+
+	public UserLabelsListEditor(String name, String labelText, Composite parent) {
+		super(name, labelText, parent);
+	}
+
+	@Override
+	protected String[] parseString(String string) {
+		if (string == null || string.equals("")) {
```
apupier: string.equals("") can be replaced by string.isEmpty() since Java 7

apodhrad: that's only syntax sugar ;-)

Regarding the functionality, it could be used only with non-uri components. For any endpoint we display its uri which is the only one reasonable information to be displayed.

So, we need to improve the information about what it actually does. And yeah... the error message could be better.

Regarding the ID, I'd like to remove this option. I don't think it is useful to display the IDs.

Thank you for your fast feedback. I'll do some changes and let you know once I'm done

apodhrad: And regarding the order of preferred labels, we should avoid using more items with the same component. I'll add the appropriate validation.

apodhrad:
- added information that this feature is not for endpoints
- removed option for switching to IDs (I don't see any use case for this option)
- improved error messages
- extended the validation by checking duplicate component preferences (cannot set both log.message and log.id)

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/CamelEditorPreferencePage.java
```diff
@@ -70,13 +70,6 @@ public void createFieldEditors() {
 
 		addField(this.defaultLanguageEditor);
 
-		this.preferIdAsLabelEditor = new BooleanFieldEditor(
-				PreferencesConstants.EDITOR_PREFER_ID_AS_LABEL,
-				UIMessages.editorPreferencePagePreferIdAsLabelSetting,
```
apupier: Need to remove the corresponding externalized string

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/UserLabelsListValidator.java
```diff
@@ -0,0 +1,64 @@
+/*******************************************************************************
+ * Copyright (c) 2013 Red Hat, Inc.
```
apupier: 2013 --> 2016

apodhrad:
done

apupier:
testPR

lhein:
Andrej, I really like the idea. I would like to add some wishes / suggestions.

1. make the list allow multi selection. this will make it easier to delete more entries at once. For the edit just always take the first object in the selection.

2. for the dialog I would like to see a more advanced version. There should be a combo box which lets you choose an EIP. (a list of available EIP's can be obtained by calling https://github.com/fusesource/fuseide/blob/master/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactory.java#L97 with parameter returned by https://github.com/fusesource/fuseide/blob/master/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactory.java#L120. That will give you a CamelModel object (https://github.com/fusesource/fuseide/blob/master/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactory.java#L120) which you can use to obtain a list of EIPs.
The EIP object itself will provide you a list of available parameters (https://github.com/fusesource/fuseide/blob/master/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactory.java#L120). So its basically filling the first combo box with all available EIPs and then register a selection listener to that box which then fills the next combo box with all parameters from that eip. If you really want to make it awesome you would have a text field too which takes whatever you select in the parameter combo box and adds it to the text box so you could create a custom label like "log.id (log.message)" :) And if you want to make it even nicer you could also add a type completion and validation in the text field. We will happily assist you with all of that if you need us.

3. the list in the pref page should be a table with 2 columns. first column will display the EIP name and the second one will display the label expression.

4. make it work with endpoints (from / to) too. 

Good work so far, Andrej!

apodhrad:
Implemented 1, 2 and 3. Regarding the last suggestion I have some doubts about that. Endpoints contain only one parameter which is 'uri'. This 'uri' may contains some attribute and I think this is out of the scope of this issue.

I will add more tests and maybe do some refactoring... In the meanwhile you can check the updated PR

lhein:
Will give it a try soon. Thanks for your help!

lhein:
Some suggestions:
# if you enable the combo boxes to be editable as it is atm then you should also select a ModificationListener to fill the second combo. If not then you should create the combo as READ_ONLY. If you keep it editable it would be nice if it autocompletes / filters with the chars you already entered, otherwise searching/scrolling this long list is cumbersome.

# the EIP combo box contains also dataformats and languages (which are never displayed in the diagram) and the from/to eips which are not supported for customizing the label. We should remove those entries from the result set. An easy approach to do so is by calling 
https://github.com/fusesource/fuseide/blob/master/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/provider/ProviderHelper.java#L161
for each EIP and whenever you receive a value not equal to "NONE" you are safe to fill it into the list. Its a bit of a dirty approach but it will do for now.

# the parameter combo contains all parameters which doesn't make sense. We should filter out all parameters which are not of kind "attribute".

# for the parameter combo the same applies as for the eip combo. If editable we should filter on what you are typing to reduce the resultset. If not editable then you should make it READ_ONLY. To be honest I would prefer editable for both combo boxes. 
See http://stackoverflow.com/questions/25241236/how-to-search-for-required-elements-in-list-using-jface for how to filter lists.

Besides that it looks really good. Maybe increase the height of the 2 column table a bit. It looks a bit tiny. 

Keep on the nice work!

Lars

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/PreferredLabelDialog.java
```diff
@@ -0,0 +1,210 @@
...
+		container.setLayout(layout);
+
+		createComponentComboBox(container);
+		createParameterCombo(container);
+
+		CamelModel model = CamelModelFactory.getModelForVersion(CamelModelFactory.getLatestCamelVersion());
+		List<String> componentNames = new ArrayList<String>();
+		model.getEipModel().getSupportedEIPs().stream().map(eip -> eip.getName()).sorted().forEach(componentNames::add);
+
+		componentCombo.setItems(componentNames.toArray(new String[componentNames.size()]));
```
apupier: you should better use a ComboViewer.
As input, you will provide the Eip.
It will be easier to manipulate then, for instance l.177, no more need to do a model.getEipModel().getEIPByName(componentName).getParameters() you will be able to directly ge the parameters.
Between the componentCombo and the parameterCombo, we can use a Master-Detail pattern for Databinding.

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/PreferredLabelEditor.java
```diff
@@ -0,0 +1,424 @@
...
+
+		table = getTableControl(parent);
+		gd = new GridData(GridData.FILL_HORIZONTAL);
+		gd.verticalAlignment = GridData.FILL;
+		gd.horizontalSpan = numColumns - 1;
+		gd.grabExcessHorizontalSpace = true;
+		table.setLayoutData(gd);
+
+		buttonBox = getButtonBoxControl(parent);
+		gd = new GridData();
```
apupier: avoid to reuse the same variable

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/PreferredLabelEditor.java
```diff
@@ -0,0 +1,424 @@
...
+
+	@Override
+	protected void doFillIntoGrid(Composite parent, int numColumns) {
+		Control control = getLabelControl(parent);
+		GridData gd = new GridData();
+		gd.horizontalSpan = numColumns;
+		control.setLayoutData(gd);
+
+		table = getTableControl(parent);
+		gd = new GridData(GridData.FILL_HORIZONTAL);
```
apupier: you can use also GridDataFactory.fillDefaults().span(numColumns -1, 1).grab(true, false).create()

(personally I prefer, and while writing it it is faster, but not sure it is worthy to change it in existing code)

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/PreferredLabelEditor.java
```diff
@@ -0,0 +1,424 @@
...
+	}
+
+	protected void packColumns() {
+		for (TableColumn col : table.getColumns()) {
+			col.pack();
+		}
+	}
+
+	protected String[] getColumnHeaders() {
+		return new String[] { "Component", "Parameter" };
```
apupier: Externalized String

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/PreferredLabelEditor.java
```diff
@@ -0,0 +1,424 @@
...
+	 * @return the combined string
+	 * @see #parseString
+	 */
+	private String createList(TableItem[] rows) {
+		return String.join(AbstractCamelModelElement.USER_LABEL_DELIMETER, getPreferredLabels());
+	}
+
+	public List<String> getPreferredLabels() {
+		List<String> preferredLabels = new ArrayList<>();
+		Stream.of(table.getItems()).map(row -> row.getText(0) + "." + row.getText(1)).forEach(preferredLabels::add);
```
apupier: I think that you don't need to create a new List and replace .forEach(preferredLabels::add) by .collect(Collectors.toList())

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/PreferredLabelEditor.java
```diff
@@ -0,0 +1,424 @@
...
+		return preferredLabels;
+	}
+	
+	public String[] getComponents() {
+		List<String> components = new ArrayList<>();
+		Stream.of(table.getItems()).map(row -> row.getText(0)).forEach(components::add);
```
apupier: I think that you can use directly .toArray(String[]::new)

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/ComponentValidator.java
```diff
@@ -0,0 +1,54 @@
...
+	public ComponentValidator(String... components) {
+		this(Arrays.asList(components));
+	}
+
+	@Override
+	public String isValid(String newText) {
+		if (newText.isEmpty()) {
+			return UIMessages.preferredLabels_errorMessageEmptyComponent;
+		}
+		if (newText.matches(".*[\\W&&[^-]].*")) {
```
apupier: I cannot find the rules in Apache Camel specification for Component and parameter name, can you point it to me?

apupier:
Can you make the width of the Component and parameter larger?
image

(ideally fill the space but this is a tricky part of SWT and sometimes complicated depending on parenting layout)

apupier:
I have this NPE when I click on "Add..."

!ENTRY org.eclipse.ui 4 0 2016-07-11 09:52:04.081
!MESSAGE Unhandled event loop exception
!STACK 0
java.lang.NullPointerException
	at org.fusesource.ide.camel.editor.preferences.PreferredLabelDialog.setWarning(PreferredLabelDialog.java:99)
	at org.fusesource.ide.camel.editor.preferences.PreferredLabelDialog.validate(PreferredLabelDialog.java:83)
	at org.fusesource.ide.camel.editor.preferences.PreferredLabelDialog$2.modifyText(PreferredLabelDialog.java:149)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:179)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4362)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1113)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1137)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1118)
	at org.eclipse.swt.widgets.Combo.setItems(Combo.java:2113)
	at org.fusesource.ide.camel.editor.preferences.PreferredLabelDialog.createDialogArea(PreferredLabelDialog.java:125)
	at org.eclipse.jface.dialogs.TitleAreaDialog.createContents(TitleAreaDialog.java:161)
	at org.eclipse.jface.window.Window.create(Window.java:430)
	at org.eclipse.jface.dialogs.Dialog.create(Dialog.java:1096)
	at org.fusesource.ide.camel.editor.preferences.PreferredLabelDialog.create(PreferredLabelDialog.java:105)
	at org.eclipse.jface.window.Window.open(Window.java:792)
	at org.fusesource.ide.camel.editor.preferences.PreferredLabelEditor.addPressed(PreferredLabelEditor.java:308)
	at org.fusesource.ide.camel.editor.preferences.PreferredLabelEditor$3.widgetSelected(PreferredLabelEditor.java:292)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:248)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4362)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1113)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4180)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3769)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:827)
	at org.eclipse.jface.window.Window.open(Window.java:803)
	at org.eclipse.ui.internal.dialogs.WorkbenchPreferenceDialog.open(WorkbenchPreferenceDialog.java:211)
	at org.eclipse.ui.internal.OpenPreferencesAction.run(OpenPreferencesAction.java:63)
	at org.eclipse.jface.action.Action.runWithEvent(Action.java:473)
	at org.eclipse.jface.action.ActionContributionItem.handleWidgetSelection(ActionContributionItem.java:595)
	at org.eclipse.jface.action.ActionContributionItem.access$2(ActionContributionItem.java:511)
	at org.eclipse.jface.action.ActionContributionItem$5.handleEvent(ActionContributionItem.java:420)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4362)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1113)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4180)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3769)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$4.run(PartRenderingEngine.java:1127)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:337)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1018)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:156)
	at org.eclipse.ui.internal.Workbench$5.run(Workbench.java:694)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:337)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:606)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:150)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:139)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:196)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:134)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:104)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:380)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:235)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.equinox.launcher.Main.invokeFramework(Main.java:669)
	at org.eclipse.equinox.launcher.Main.basicRun(Main.java:608)
	at org.eclipse.equinox.launcher.Main.run(Main.java:1515)
	at org.eclipse.equinox.launcher.Main.main(Main.java:1488)

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/preferences/PreferredLabelDialog.java
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	public boolean setWarning(String msg) {
+		if ((msg == null || msg.isEmpty()) && getButton(IDialogConstants.OK_ID) != null) {
+			getButton(IDialogConstants.OK_ID).setEnabled(true);
+			setMessage(null, IMessageProvider.NONE);
+			return false;
+		}
+		setMessage(msg, IMessageProvider.WARNING);
+		getButton(IDialogConstants.OK_ID).setEnabled(false);
```
apupier: 5 lines above you check for getButton(IDialogConstants.OK_ID) != null but ,not (which explains my reported NPE)

apodhrad:
Hey Lars,
regarding the 3
"the parameter combo contains all parameters which doesn't make sense. We should filter out all parameters which are not of kind "attribute"."

Actually, they make sense... for example "description". At the moment I would rather keep all parameters.

I started using ComboViewer instead of Combo but didn't implement the filtering while you are typing into the combo. I set it as readonly ;-)

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/CamelEditor.java
```diff
@@ -771,9 +771,9 @@ protected void setInput(IEditorInput input) {
 	 */
 	@Override
 	public void propertyChange(PropertyChangeEvent event) {
-		if (event.getProperty().equals(PreferencesConstants.EDITOR_PREFER_ID_AS_LABEL)) {
-			// user switched the displaytext logic flag - refresh diagram and outline
-			designEditor.update();
+		if (event.getProperty().equals(PreferencesConstants.EDITOR_PREFERRED_LABEL)) {
+			// user defined own labels
+			designEditor.update();;
```
apupier: there is an extra semicolon

apupier:
testPR

apodhrad:
Is there something what you would like to add/edit? I removed the second semicolon and rebased to the master branch so that it is ready to merge/

apupier:
testPR

apupier:
@apodhrad there is still one comment about the NPE

apodhrad:
Fixed the problem with NPE

lhein:
@apodhrad any chance u rebase ur PR with current master?

apodhrad:
@lhein yes, I will come back to this PR in 1-2 days

lhein:
@apodhrad can you rebase your PR?

apodhrad:
I'm working on it and I'm already done - still need to fix some conflicts.

apodhrad:
Unfortunately, I'm not able to change this PR since the PR refers to
```
unknown repository
```
so I created new PR at #1133

PR #567, opened 2016-07-06
FUSETOOLS-1928 - Generate Unique ID for routes in same project

apupier:
/!\ we are looking unique ID on saved model files. If one other camel
file in the same project has been modified, two ids can be identic.


core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/util/CamelFilesFinder.java
```diff
@@ -0,0 +1,65 @@
...
+	
+	public Set<IFile> findFiles(IResource resource) {
+		Set<IFile> res = new HashSet<>();
+		if (resource instanceof IContainer ) {
+			try {
+				IResource[] children = ((IContainer)resource).members();
+				for (IResource f : children) {
+					if (f instanceof IContainer) {
+						// ignore the target folder
+						if (f.getName().equalsIgnoreCase("target")
```
lhein: you should also exclude the bin folder

lhein:
+1

PR #571, opened 2016-07-07
FUSETOOLS-1954 - Fix NPE

apupier:
To be bale to Open Route from Project Explorer when editor already
opened with 'Source' or 'Configuration' tab selected


lhein:
+1

PR #572, opened 2016-07-07
FUSETOOLS-1955 - fixing NPE when switching between routes

lhein:


editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/features/custom/GoIntoContainerFeature.java
```diff
@@ -58,24 +59,32 @@ public boolean canExecute(ICustomContext context) {
...
+	private boolean isAllowedToExecute(AbstractCamelModelElement cme) {
+		// go into is allowed if:
+		// - selected element is a route
+		// - selected container is the camel context
+		// - context contains more than one route
+		// go up is allowed if:
+		// - selected container is a route
+		// - selected container is the selected element
+		CamelDesignEditor editor = (CamelDesignEditor)getDiagramBehavior().getDiagramContainer();
+		return 	(editor != null &&
```
apupier: editor != null can be factorized

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/features/custom/GoIntoContainerFeature.java
```diff
@@ -58,24 +59,32 @@ public boolean canExecute(ICustomContext context) {
...
+		// go into is allowed if:
+		// - selected element is a route
+		// - selected container is the camel context
+		// - context contains more than one route
+		// go up is allowed if:
+		// - selected container is a route
+		// - selected container is the selected element
+		CamelDesignEditor editor = (CamelDesignEditor)getDiagramBehavior().getDiagramContainer();
+		return 	(editor != null && 
+				 cme.getNodeTypeId().equals("route") &&
```
apupier: maybe we can extract each side of the ||
if I understand well th ecomment, we will have something like isGoIntoAllowedToExecute(editor, cme)|| isGoUpAllowedToExecute(editor, cme)
The condition will easier to understand, what do you think?

apupier: seems that there is already a isGoInto method which looks very close

apupier:
+1

PR #576, opened 2016-07-11
FUSETOOLS-1970 - Clean code during investigation

apupier:


lhein:
+1

bfitzpat:
Ok. I've tried with three different examples and it works fine if you go from a perspective other than the Git perspective and do a Maven project import. However, if in the Git perspective you drill into the git repo path, right click on a Java DSL fuse project and select "Import Maven Projects" from the popup menu - it somehow does not add the Camel facet still.

bfitzpat:
I'm not sure what's different about that particular triggering of the maven project import, but it does not work fully with the Camel project configurator automatically adding the camel facet.

PR #577, opened 2016-07-11
FUSETOOLS-1970 - Ensure a build has been done before searching in Java

apupier:
classpath

- also wait for Refresh Build finished to avoid some "Resource out of
sync" even if no issue detected from that, except trace in log
- also provide the monitor and declare it as Done, the creator of the
monitor is reponsible for that.

### DIFF
--- a/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/AbstractProjectTemplate.java
+++ b/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/AbstractProjectTemplate.java
@@ -11,12 +11,16 @@
 package org.fusesource.ide.projecttemplates.adopters;
 
 import org.eclipse.core.resources.IProject;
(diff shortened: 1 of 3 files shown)


apupier:
@lhein and @bfitzpat as it is an asynchronous issue, can you test it on your machine?

bfitzpat:
Ok. I've tried with three different examples and it works fine if you go from a perspective other than the Git perspective and do a Maven project import. However, if in the Git perspective you drill into the git repo path, right click on a Java DSL fuse project and select "Import Maven Projects" from the popup menu - it somehow does not add the Camel facet still.

bfitzpat:
I'm not sure what's different about that particular triggering of the maven project import, but it does not work fully with the Camel project configurator automatically adding the camel facet.

apupier:
ok, do you agree to merge this branch and open another JIRA issue to mange the import from git perspective?

bfitzpat:
+1 to that. sounds like a plan

bfitzpat:
Spun off FUSETOOLS-1973

PR #579, opened 2016-07-11
FUSETOOLS-1953 - Better filtering to avoid logged exception and

apupier:
initialization error

- avoid to trigger XML Camel Route validation on all xml files
- avoid to check for content types when file is out of sync

### DIFF
--- a/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/util/CamelFilesFinder.java
+++ b/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/util/CamelFilesFinder.java
@@ -69,6 +69,7 @@ public class CamelFilesFinder {
 	 */
 	public boolean isFuseCamelContentType(IFile ifile) throws CoreException {
 		return ifile != null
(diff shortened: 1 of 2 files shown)


core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/util/CamelFilesFinder.java
```diff
@@ -69,6 +69,7 @@ private boolean isWorkProjectFolder(IProject project, IResource f) {
 	 */
 	public boolean isFuseCamelContentType(IFile ifile) throws CoreException {
 		return ifile != null
+				&& ifile.isSynchronized(IResource.DEPTH_ZERO)
```
lhein: shouldn't we just refresh if we find a file out of sync and then do the check?

apupier: I was also wondering for that.
On the other hand, I don't find that this part of the code should be responsible for the refresh. It can mean that there are still actions processing and that the file might change again. This code will be called again after the refresh is finished.

lhein:
+1

PR #582, opened 2016-07-12
FUSETOOLS-1939 - Check for Connection to avoid UI freeze

apupier:
If the connection is lost (network died or local process killed), the UI
is no more frozen when clicking in JMX Navigator (or only very few
seconds)
The Node "maven" vanishes after few seconds.


lhein:
+1

PR #589, opened 2016-07-18
FUSETOOLS-1915 - fixes NPE in console during validation bundle shutdown

bfitzpat:


bfitzpat:
I'm not sure why this chose to replace the whole file and not just the one affected line. I can delete the PR and start over to see if it fixes it.

lhein:
maybe wrong settings for line endings or tabs/spaces.

bfitzpat:
I've tried setting my local git to - git config --global core.autocrlf true

And that didn't make a difference.

PR #592, opened 2016-07-20
FUSETOOLS-1990 - only install our needed facets if its a camel projec…

lhein:
…t (contains camel context)


apupier:
there is no test for such a critical issue?

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/maven/CamelProjectConfigurator.java
```diff
@@ -68,7 +68,9 @@ public void mavenProjectChanged(MavenProjectChangedEvent event, IProgressMonitor
 			IProject project = facade.getProject();
 			MavenProject mavenProject = facade.getMavenProject(monitor);
 			IFacetedProject fproj = ProjectFacetsManager.create(project);
-			installDefaultFacets(project, mavenProject, fproj, monitor);
+			if (checkCamelContextsExist(project)) {
```
apupier: I just realized that it means that we will slow down every Maven import because we will wait for a complete build to have finished in order to search for Java possible file extending RouteBuilder.
I have no solution yet... (and it will be always better than a NPE and isntalling Camel facets on a project which is not related^^)
But we need to start thinking about it

apupier:
I'm working on writing the test case

apupier:
please look at https://github.com/fusesource/fuseide/pull/592

PR #593, opened 2016-07-20
FUSETOOLS-1870 - Provide more Integration tests for sample projects

apupier:
- test for 2.15.1 and 2.17 redhat versions
- test Content Based Router samples
- test Launch in Debug mode for Spring and Blueprint CBR samples


apupier:
The JMX connection can't be established on CI...
need to check if port is not occupied
need to check if the Launch of the "Debug as local camel context" is working on this Linux installation...

apupier:
testPR

apupier:
testPR

apupier:
testPR

apupier:
the PR job is not triggered and it seems to be stuck, I will open another PR

PR #595, opened 2016-07-21
FUSETOOLS-1993 - restrict the choice to a single otherwise

lhein:


apupier:
we need some tests otherwise we will jump into this regression again

apupier:
+1 for this.
We shoudl ask QE to wirte also a UI automated test for DnD behavior on "otherwise" in choices

PR #600, opened 2016-07-22
FUSETOOLS-1997 - fix for weird NPE in uninstall of karaf bundle

bfitzpat:


apupier:
+1 we need to ask QE to ensure that they have test with a Fuse Runtime Server on this kind of project.
Or try to have this kind of integration test in our own test suite? In this case, maybe worth creating a separated JIRA task.

bfitzpat:
Added a note to the JIRA to see what Andrej thinks

PR #601, opened 2016-07-25
FUSETOOLS-1865 - Implement Monitors when creating new Project

apupier:
- not cancellable
- Show progress

I have currently no idea how to write automatic test for monitors correctly...


editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/AbstractProjectTemplate.java
```diff
@@ -47,18 +48,18 @@ public boolean supportsDSL(CamelDSLType type) {
 	 * @return	true on success, otherwise false
 	 * @throws CoreException	on any error
 	 */
-	public final boolean create(IProject project, NewProjectMetaData projectMetaData) throws CoreException {
-		IProgressMonitor monitor = new NullProgressMonitor();
+	public final boolean create(IProject project, NewProjectMetaData projectMetaData, IProgressMonitor monitor) throws CoreException {
+		SubMonitor subMonitor = SubMonitor.convert(monitor, 100);
```
jpav: I'm pretty new to the SubMonitor, but seems like you use this differently in the rest of this PR.  The division of the work into arbitrary chunks (e.g., 30 vs. 15, etc.) seems just that, arbitrary.  Would it not be better to use the approach you take in other areas of the code and just give each individual task a "1"?  Maybe I'm missing something though...

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/configurators/MavenTemplateConfigurator.java
```diff
@@ -111,29 +108,28 @@ private void waitAllJobsComplete(IProgressMonitor monitor) {
...
 			MavenUtils.updateCamelVersionPlugins(m2m.getBuild().getPlugins(), camelVersion);
-
+			subMonitor.worked(1);
 			OutputStream os = new BufferedOutputStream(new FileOutputStream(pomFile));
 		    MavenPlugin.getMaven().writeModel(m2m, os);
-			IFile pomIFile2 = project.getProject().getFile("pom.xml");
+			IFile pomIFile2 = project.getProject().getFile("pom.xml"); //$NON-NLS-1$
 			if (pomIFile2 != null) {
-				pomIFile2.refreshLocal(IResource.DEPTH_INFINITE, monitor);
+				pomIFile2.refreshLocal(IResource.DEPTH_INFINITE, subMonitor.newChild(1));
```
jpav: A question rather than a comment: Would the monitor handle things correctly if you only allocated 4 in the convert, but this if statement ended up adding a fifth segment?

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/AbstractProjectTemplate.java
apupier: The division of work should reflect the proportion.
In other places, there are often long-running tasks for which I wasn't able/don't take time to guess/determine the proportions, so I just put '1' for each of them.
I tried to be more precise at this place

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/configurators/MavenTemplateConfigurator.java
apupier: AFAIK, in this case, you will just have visually the progress bar at the end before the real end.
There won't be errors but from a usability point of view this is not nice.

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/AbstractProjectTemplate.java
jpav: I understand, but maybe something to consider is that I think most users are pretty used to the individual chunks of a progress monitor not even coming close to reflecting an actual time.  Meaning, for instance, given a monitor where it visually appears to have 3 equally sized tasks, each taking up a third of the bar, the first and second may for instance be very quick, while the third takes in actuality 90% of the time.  Even that, seems like it might make more sense to NOT guess at some “in the ballpark” proportion and just set everything to have the same number of ticks.  Just something to consider.

bfitzpat:
Monitors and submonitors are always a tricky business, but I think this looks ok to me. As @jpav has pointed out, they've never been exactly ideal. But they do work to give at least some indication that work is progressing. So I'm +1 with this change.

PR #604, opened 2016-07-26
Fusetools 1936 maven compilerlevel incompatible upgrade to java8

apupier:
I think that the best is to put compiler version to 1.8 in the pom.xml as 1.7 is EOL for public and we are free to provide samples that we want :-)
There is no error mentioned anyway and for me it is more to m2e/Java facets to deal with issues between maven compiler settings and java project facet.

Please review and merge this PR before: https://github.com/fusesource/fuseide/pull/603


bfitzpat:
Looks good. We should also handle the compiler level in the other samples maybe?

apupier:
Yes I think so too.
I wanted another poitn of view before doing it

PR #609, opened 2016-08-02
FUSETOOLS-1581 - Remove Breakpoint and Debug listeners when editor is

apupier:
disposed

I don't see easy and meaningful way to test it, there is no method to know if the listener are still registered or not.


bfitzpat:
+1 looks like a logical way to handle the listener.

PR #613, opened 2016-08-03
FUSETOOLS-1908 - basic User Guidance for Java DSL local launch

apupier:
just provide a message dialog to explain how user can launch the Camel
route instead of just silently ignore it.

image

Suggestions to improve the message? Do I miss some use cases?

The best would be to create the correct Maven launch configuration (or Camel Launch configuration) but I think that it can be done in a second iteration


bfitzpat:
Ok, I see what you're getting at here but I think this is misleading. Do we have any better options? 

The message does explain what needs to be done, but doesn't suggest a way to do it in the tooling. And what happens if it's not a Java DSL "sample" - but a customer's project without a readme? 

This seems very incomplete without some way to tell them to actually execute the project in the tooling. Are we sure we can't add a new "Run As...->Maven Exec Java" option (looks like the rest of the Maven options are contributed directly by m2e)? Or, better yet, update the Run As->Local Camel Context options to simply do it behind the scenes?

apupier:
This popup is triggered only when we use the "Run as local camel context". This menu is activated only on Camel projects: they are always maven ones and m2e is always available.

Users using Java DSL are supposed to be Advanced Technical users so I thought that just telling them to use mvn exec:java was enough to let them use the m2e launch configuration. Maybe I should precise how to do it:[...]it can be locally launched using "Run as -> Maven build..." and specifying goal exec:java.[...]

"And what happens if it's not a Java DSL "sample" - but a customer's project without a readme?"
If it is a Java DSL but not a sample, I suggest to have a look to the sample to see how it is working :-)

"Are we sure we can't add a new "Run As...->Maven Exec Java" option (looks like the rest of the Maven options are contributed directly by m2e)?"
maybe but I'm not sure that we will be able to provide the menu only at the right-time with reasonable performance. I see that with less priority than other items planned for 8.0.0

"Or, better yet, update the Run As->Local Camel Context options to simply do it behind the scenes?"
Yes this is exactly what I want but I propose it in a second iteration. This really not the same amount of work and with less priority than other items planned for 8.0.0

apupier:
updated message:
image

bfitzpat:
That's an improvement. :) And I certainly get the priority as being lower for this release. Here's a suggested change in the wording:

"No XML-based Camel context file has been located in the project you are trying to execute. If the project is Java DSL-based, you can execute locally using "Run As->Maven build..." menu and specifying the "exec:java" goal. Consult the documentation for additional details."

Slight adjustment to the text except for the last line. I think this is something we'll need to get into the documentation via Jane. Pointing people to look at project templates (we're not calling them samples, are we) may require that they shift gears completely to not only create a new project and then take a guess at how running the template may work with their project. Too much ambiguity. 

In the Help that's in there we already have the topic "Red Hat JBoss Fuse Tooling for Eclipse > Using the Red Hat JBoss Fuse Tooling for Eclipse Tools > Developing Applications > Running routes inside Red Hat JBoss Fuse Tooling". We should simply have Jane add some text around testing Java DSL routes and I think we'll be good.

bfitzpat:
We could even say "Consult the Online Help for additional details about Running routes using Maven." to be even more specific.

apupier:
updated text:

image

bfitzpat:
+1

bfitzpat:
@apupier  Can you create a spin-off JIRA for Jane and the docs team so we get the exec:Java bits put into the help and documentation?

apupier:
Jane was faster than me: https://issues.jboss.org/browse/FUSEDOC-1514

PR #614, opened 2016-08-03
FUSETOOLS-2002 - Drop Apache ServiceMix support

apupier:
- left embedded Documentation references which will be updated later on

there is also one file with service mix references which seems not used: archetypes-catalog.xml
I removed it, everything seems to work fine.


lhein:
+1

PR #616, opened 2016-08-04
FUSETOOLS-1817 - Convert 'To' To 'From' if endpoint is the first added

apupier:


bfitzpat:
+1 - makes perfect sense to make this easier for the user

PR #617, opened 2016-08-04
FUSETOOLS-1997 - Have deployment before build working

apupier:
- call mvn package before each full publication of a Module.

the downside of this approach is that it takes more time to do a full publish BUT it forces to have an up-to-date version when we call "full publish" I think that it is what user wants

note: We don't have test with a Fuse Runtime Server, currently only QE does that


bfitzpat:
+1 So long as everything we need to actually publish to the server is generated, I think people will be ok with a bit more time to build.

PR #619, opened 2016-08-05
FUSETOOLS-2019 - Support Local CamelContext Launch on source tab

apupier:


bfitzpat:
+1

lhein:
-1 - this behaviour will confuse users. If they invoke the execute on the editor they will expect that the current editor content will be used for execution but in fact all unsaved changes are not considered and skipped out of the execution. For a launch from inside the editor we should save a temporary file somewhere and launch that one instead of the file in the project OR alternatively simply remove the execute option from the source tab of the editor.

apupier:
I think this is less confusing to have a launch started than ignoring the UI action completely.
We can also pop a dialog to warn him and ask him to save the editor. I feel that it should be done in another PR.

lhein:
yeah, thats not what I suggested. I suggested to either remove the action from the editor OR to do it correctly like described in my last comment. But adding a warning dialog in case of a dirty editor might be a good starting point and should be part of this PR.

apupier:
if the editor is not saved, the current behavior is to ask user to save resources:
image

lhein:
Ok, then all is fine I'd say.

PR #621, opened 2016-08-08
FUSETOOLS-2022 - Fix Regression And Add Test

apupier:
- fix regression introduced with FUSETOOLS-1987
- add tests to ensure loading of xml currently provided by JMX is
correctly loaded


lhein:
+1

PR #622, opened 2016-08-08
FUSETOOLS-1875 - respect disablement of template description

apupier:
- use Text instead of StyledText
- remaining little issue for Search icon place on filter Text but should
be an SWT issue


apupier:
enabled:
image

disabled:
image

lhein:
+1

PR #623, opened 2016-08-09
FUSETOOLS-1979 - Allow update dependencies of custom components

apupier:
When Camel version is updated, custom components might want to update
their dependencies too


lhein:
+1

PR #627, opened 2016-08-10
FUSETOOLS-1970 - Improve robustness during investigation

apupier:
- ensure hidden project is open
- ensure file is not out of sync (to avoid stacks in log from
Validation)


lhein:
+1

PR #634, opened 2016-08-17
FUSETOOLS-1669: New Transformation Wizard - XML, JSON - search button shows schema files and instance documents together

jpav:


transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/XMLPage.java
```diff
@@ -331,7 +331,7 @@ private String selectResourceFromWorkspace(Shell shell, final String extension)
         }
         ClasspathResourceSelectionDialog dialog = null;
         HashSet<String> extensions = new HashSet<>();
-        extensions.add("xml"); //$NON-NLS-1$
+        if (!"xsd".equals(extension)) extensions.add("xml"); //$NON-NLS-1$
```
lhein: I think it would make sense to create constants for those values. (xml, xsd, wsdl)

jpav: IMO, these are true constants, not "magic" values that would benefit from a defined constant in case they changed.  Adding constants for these just adds extra code with no real benefit to maintenance, code size, etc.

lhein:
There was no +1 on this PR. Why did u merge it?

jpav:
I guess I misunderstood the protocol then.  I thought the +1 was for when there were no comments.  I assumed if i addressed the comments, it was okay to merge.  So now I take it that you want to wait for a +1 to the comments, too, or I guess another PR to be submitted addressing them followed by a +1?

lhein:
The intended flow is to wait for review of your PR. This will either:

end in getting comments to improve the code which can evolve into a discussion if needed 
and at some point if all are satisfied with the result will result in a +1 or -1 if not.

OR 

you could get a +1 directly. If you got a +1 and this PR is at least ~1 day old (that's the time we wanted to take for a proper review) then you or the reviewer can merge the PR.

OR

you could get a -1 directly but then with a good explanation why you decline the PR.

Does that make sense?

jpav:
K, that’s a bit different than I thought, but I get what you’re explanation below.  This is kind of a separate issue from this event, but I thought the day wait time was in lieu of getting a comment or +1, i.e., a time period to wait until a review was done or, if not, assume it’s okay to check in.  Is that wrong?

PR #635, opened 2016-08-17
FUSETOOLS-1537: ClassNotFoundException if you use json/xml file which contains underscore in Data Transformation

jpav:


transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/XMLPage.java
```diff
@@ -331,7 +331,7 @@ private String selectResourceFromWorkspace(Shell shell, final String extension)
         }
         ClasspathResourceSelectionDialog dialog = null;
         HashSet<String> extensions = new HashSet<>();
-        extensions.add("xml"); //$NON-NLS-1$
+        if (!"xsd".equals(extension)) extensions.add("xml"); //$NON-NLS-1$
```
apupier: need to add //$NON-NLS-2$

apupier: personally, I prefer to have it wirtten on 2 lines with brackets

apupier:
better to create a separate branch based on master, it will avoid mixing commits. They are not requiring each other.

apupier:
A test is needed to ensure that no regression on the '_' case..

jpav:
Crap, I didn't realize I had built this branch from the other.  I'll fix that and address the other comments.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/wizards/NewTransformationWizard.java
```diff
@@ -434,14 +432,16 @@ private String generateModel(String filePath,
             return getJavaQualifiedName(filePath, monitor);
         }
         case JSON: {
+            String className = classNameBuilder.toString().replace("_", "");
```
apupier: Why '_' is a special character here? Is it possible to have issues with some other characters?

jpav: The 3rd-party library we're using is removing the underscore from the name during code generation.  Not sure what other characters might be treated the same way, or how to even determine this.  It will just take some experimentation I think.

jpav: After looking at our tests, we don't have enough in the test framework to test this; the code is on the UI side only and is not conducive to allowing for tests right now.  I already have an issue logged to address our lack in testability in the UI, so this will have to be addressed when that issue is addressed.

apupier: Rethinking about it after your comment. I think that JsonModelGeenrator should be responsible to manage the special characters and in this case it will be easier to test the JSonModelGenerator

apupier: "The 3rd-party library we're using is removing the underscore from the name during code generation. Not sure what other characters might be treated the same way, or how to even determine this. It will just take some experimentation I think"
we can have a look to source code of used library which seems to be https://github.com/joelittlejohn/jsonschema2pojo or by experimentation we can write a paraterized test providing all (java or UTF-8 or ??) possible characters and checking for the output

jpav: Regarding the last comment, yes, the test you suggest would probably be the way to go.  Regarding the model generator comment, that's a good idea.  That will also allow me to write a regression test for this right now.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/Hl7Page.java
```diff
@@ -76,8 +76,8 @@
      */
     public Hl7Page(final String pageName, final Model model, final boolean isSource) {
         super(pageName, model);
-        setTitle("HL7 Page");
-        setImageDescriptor(Activator.imageDescriptor("transform.png"));
+        setTitle("HL7 Page"); //$NON-NLS-1$
```
apupier: better to externalize the title

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/Hl7Page.java
```diff
@@ -199,11 +199,11 @@ private void updateSettingsBasedOnFilePath(String path) throws Exception {
     @Override
     public void createControl(final Composite parent) {
         if (this.isSource) {
-            setTitle("Source Type (JSON)");
-            setDescription("Specify details for the source JSON for this transformation.");
+            setTitle("Source Type (JSON)"); //$NON-NLS-1$
```
apupier: better to externalize the title

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/Hl7Page.java
```diff
@@ -199,11 +199,11 @@ private void updateSettingsBasedOnFilePath(String path) throws Exception {
...
+            setDescription("Specify details for the source JSON for this transformation."); //$NON-NLS-1$
```
apupier: better to externalize the description

jpav: Please ignore all of these in the HL7Page.  This is just a placeholder right now and is not used by the code.  It was created a while ago in preparation for when we have first-class HL7 support.  The code will change significantly then, so it’s not worth addressing now.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/wizards/NewTransformationWizard.java
jpav: Well, that didn't turn out well at all.  it turns out we need the generated class name for subsequent use by the tooling, but the generator not only doesn't return that directly, but doesn't provide any way that I can discern to distinguish the actual top-level class we want to map from the other top-level classes it generates, at least not with some significant additional processing and code.  So, for the time being I'm going to leave the code as is with the underscore being dealt with in the wizard.  The worst impact of this is that until I figure out a way to determine the class to be mapped from the generator, I can't effectively produce a good regression test for this either.

jpav:
I've updated the code concerning the NLS comments, so looking for a +1 now...

bfitzpat:
+1

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/wizards/NewTransformationWizard.java
apupier: I fear to understand that with the current state, in the wizard there will be no error, some code will be generated but this code will not be used by the tooling after. So user, will be blocked in a further step? If it is the case, it seems even worse from a functional point of view. I would prefer to have a popup explaining to user that he can't use what he selected because there are underscores in it.

PR #637, opened 2016-08-17
FUSETOOLS-2040 - Improve definition of endpoints

apupier:
- avoid issues when something other than an endpoint have a parameter
named "uri"

I checked the other places , the isEndpointElement is called and in fact it was leading to issues too. So other parts are fixed too :-)

### DIFF
--- a/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
+++ b/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
@@ -40,6 +40,8 @@ import org.w3c.dom.NodeList;
  */
 public abstract class AbstractCamelModelElement {
 
+	static final String URI_PARAMETER_KEY = "uri";
+	private static final String WIRE_TAP_NODE_NAME = "wireTap";
 	public static final String ENDPOINT_TYPE_TO = "to";
 	public static final String ENDPOINT_TYPE_FROM = "from";
 	public static final String TOPIC_REMOVE_CAMEL_ELEMENT = "TOPIC_REMOVE_CAMEL_ELEMENT";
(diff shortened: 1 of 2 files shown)


core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
```diff
@@ -50,7 +52,7 @@
 	public static final String CHOICE_NODE_NAME = "choice";
 	public static final String WHEN_NODE_NAME = "when";
 	public static final String OTHERWISE_NODE_NAME = "otherwise";
-	public static final String WIRETAP_NODE_NAME = "wireTap";
+	public static final String WIRETAP_NODE_NAME = WIRE_TAP_NODE_NAME;
```
lhein: can't we just remove one of the 2 constants?

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
```diff
@@ -438,7 +440,12 @@ public AbstractCamelModelElement getInputElement() {
 	}
 
 	public boolean isEndpointElement() {
-		return getParameter("uri") != null && ((String) getParameter("uri")).trim().length() > 0;
+		if(Arrays.asList(ENDPOINT_TYPE_FROM, ENDPOINT_TYPE_TO, ENDPOINT_NODE_NAME).contains(getNodeTypeId())){
+			Object uriParameter = getParameter(URI_PARAMETER_KEY);
+			return uriParameter != null && ((String) uriParameter).trim().length() > 0;
```
lhein: an endpoint is an endpoint....not important if uri is filled or not?

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
```diff
@@ -1559,18 +1566,6 @@ public String getDocumentationFileName() {
 	}
 
 	/**
-	 * returns the category this item belongs to
-	 *
-	 * @return
-	 */
-	public String getCategoryName() {
```
lhein: why did u remove that?

apupier: it wasn't used

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
apupier: maybe, I kept this part of the old definition. I'm not sure about edge case that it can imply to remove it.

lhein:
I still would remove the check for uri being non-empty and non-null. The check for the nodetypeid should do the trick.
Besides that +1.

lhein:
looks good here! +1

PR #641, opened 2016-08-18
ENTESB-5845: make activemq templates working on runtime

lburgazzoli:


apupier:
Can you rename the commit message which is currently three times repeated same sentence?
(if you can rebase on master at the same time, it will be perfect)

lburgazzoli:
@apupier done

apupier:
@lburgazzoli thx.
I created a PR on your branch based on the comments that i previously done on the commit itself, see https://github.com/lburgazzoli/fusesource-fuseide/pull/1
after that I think we can merge this PR.
I'm still a bit worried about the need to create a fiel in Fuse Runtime itself but we will see how to make it user-friendly in a second time

apupier:
+1 for integrating in code base then I will work on another PR to integrate it in Fuse Tooling Template framework

apupier:
+1
To integrate fully the AMQ templates, there is still some work to do but they are relted to Fuse Tooling itself, not the template itself

apupier:
@lburgazzoli thank you for the help to provide this template

PR #643, opened 2016-08-18
FUSETOOLS-2045 - update facets based on maven packaging type

bfitzpat:


bfitzpat:
I have no clue why this has created an infinite loop in the test.

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/maven/CamelProjectConfigurator.java
```diff
@@ -156,9 +160,20 @@ private void installDefaultFacets(IProject project, MavenProject mavenProject, I
...
-			installFacet(fproj, fpwc, javaFacet, javaFacet.getLatestVersion(), monitor);
-			installFacet(fproj, fpwc, m2eFacet, m2eFacet.getLatestVersion(), monitor);
-			installFacet(fproj, fpwc, utilFacet, utilFacet.getLatestVersion(), monitor);
+			
+			// adjust facets we install based on the packaging type we find
+			String packaging = null;
+			if (mavenProject.getPackaging() != null) {
+			    packaging = mavenProject.getPackaging().toUpperCase();
+			}
+            installFacet(fproj, fpwc, javaFacet, javaFacet.getLatestVersion(), monitor);
```
apupier: seems that there is an issue with indentation

apupier:
Need to provide an IT test for all templates, please see FuseIntegrationProjectCreatorRunnableIT and the implementation classes.

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/maven/CamelProjectConfigurator.java
```diff
@@ -156,9 +160,20 @@ private void installDefaultFacets(IProject project, MavenProject mavenProject, I
...
+            installFacet(fproj, fpwc, m2eFacet, m2eFacet.getLatestVersion(), monitor);
+			if (packaging != null && WAR_PACKAGE.contentEquals(packaging)) {
```
lhein: the != null check seems to be overhead here. btw why do u use contentEquals instead of equals(IgnoreCase) ?

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/maven/CamelProjectConfigurator.java
```diff
@@ -156,9 +160,20 @@ private void installDefaultFacets(IProject project, MavenProject mavenProject, I
...
+			    installFacet(fproj, fpwc, webFacet, javaFacet.getLatestVersion(), monitor);
+			} else if (packaging != null && (BUNDLE_PACKAGE.contentEquals(packaging)
```
lhein: see above

apupier:
testPR

apupier:
> I have no clue why this has created an infinite loop in the test.

I relaunched the build and it works fine, seems that we have a robustness issue here.

bfitzpat:
Updated the PR to address the indention and equalsIgnoreCase questions from earlier. I left the check for != null because I am a paranoid person and want to make sure. :) We'll see if the test PR passes this time.

Thanks for the feedback. Want to get this one in so I can go back to the eap template and see why it's almost but not quite working.

lhein:
+1

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/maven/CamelProjectConfigurator.java
```diff
@@ -152,17 +156,29 @@ private void configureFacet(MavenProject mavenProject, IProject project, IProgre
...
-			configureNature(project, mavenProject, monitor); 
-		}
+	        throws CoreException {
+	    String camelVersion = getCamelVersion(mavenProject);
+	    if (camelVersion != null) {
+	        IFacetedProjectWorkingCopy fpwc = fproj.createWorkingCopy();
+
+	        // adjust facets we install based on the packaging type we find
+	        String packaging = null;
+	        if (mavenProject.getPackaging() != null) {
```
apupier: any reason why the packaging is calculated here? and not just before we need it?

If no, what do you think of grouping them and perhaps even extracting a method for configureFacetDependingOnFacet?

bfitzpat: Interesting question. I honestly don't know enough about how/when certain things are fired in the maven framework to know one way or another what would be better. @lhein what do you think? Can we address this in a different JIRA?

lhein: I can't see what you mean Aurelien. Could you be more specific what you want to change?

apupier: the packaging variable is used only at line 171.
We have two lines installFacet which are nto related to installFacet

If we put the packaging affectation closer to l.171, we can see that we have a kind of group of conditions that are here for a single reason configureFacetDependingOnFacet

lhein:
That is a quite minor change. Brian/Aurelien, can you issue a new PR for this change? We don't need to track it in another Jira.

PR #645, opened 2016-08-19
FUSETOOLS-1977 - Include AMQ Template

apupier:
- provide Spring and Blueprint AMQ templates
- provide IDs to every elements
- align fuse version to camel version when no runtime is set at Template
creation
- Provide IT tests for Spring and Blueprint AMQ templates

/!\ AMQ Blueprint with camel 2.15.1.redhat-621084 is not working due to
https://issues.jboss.org/browse/FUSETOOLS-1986


core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactory.java
```diff
@@ -176,4 +179,12 @@ public static String getCamelVersionFromMaven(IProject project) {
         }
         return null;
 	}
+
+	public static String getFuseVersionForCamelVersion(String camelVersion) {
+		if(camelVersion.startsWith(CAMEL_VERSION_2_15)){
+			return FUSE_VERSION_6_2_1_REDHAT_084;
+		} else {
+			return FUSE_VERSION_6_3_0_REDHAT_077;
```
lhein: I'd rather not include build numbers in the names of the constants as its rather unlikely we have more than one camel version for a specific runtime.

apupier: they are the values that we will put in the pom.xml, I think that we need the whole numbers

lhein: Just be warned that we need to change the name of the constant then once the final cut of Fuse 6.3 is done.

lhein:
+1 besides the mentioned rename issue

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactory.java
apupier: yes I know but I have no ideas how to do otherwise

PR #649, opened 2016-08-22
FUSETOOLS-2044 - fixed IndexOutOfBounds Exception

lhein:


bfitzpat:
That looks like a great change to protect against unexpected consequences. +1

PR #651, opened 2016-08-23
FUSETOOLS-1987 - Avoid errors when no CamelContext and CamelOutlinePage

apupier:
view is opened

- need to modify tricky bits (SafeRunnable.setIgnoreErrors) in order to
be reproduce error in test, because eclipse is turning the bit off in
automated tests


lhein:
testPR

lhein:
+1

PR #655, opened 2016-08-23
JBIDE-23038 - Fuse runtime detection never makes a server

robstryker:


apupier:
@robstryker Can you precise the use case in order that I can test it please?

lhein:
I think that is the runtime detection mechanism. You can put the Fuse distro into the runtimes folder of RHDS/JBDS and upon restart it should automatically detect the Fuse Runtime (and create a server?).

@robstryker correct me if my assumption is wrong

robstryker:
Lars is correct. In current version, if you use the runtime detection
feature as mentioned by Lars, it will create a runtime for the given
server.

Other ways to check it are window -> preferences,  JBoss Tools/JBoss
Runtime Detection.

From here you can add a path to search and it should locate and create the
runtime in eclipse.  But it won't create the server.

Ideally, following this workflow, it should also create a server in the
server's view. Not just a runtime hidden somewhere.

apupier:
+1
yeah, that's cool, it automatically create it from New Fuse Integration Wizard now (it was something I wanted to improve)
Next step will be for us to add the newly created project directly in the "configured resources" to be deployed on the server

PR #656, opened 2016-08-24
FUSETOOLS-2058: Do not ignore any parameter in PropertiesUtilsTestIT

apodhrad:


apupier:
as note for the future, it will be nice to have this test running for all supported camel versions and not only the latest one

PR #657, opened 2016-08-26
FUSETOOLS-2053 - fix detection of Camel Version on Fuse on EAP runtimes

lhein:


apupier:
+1

apupier:
For later: this detection (for EAP and Fuse) is based on filesystem layout inside the containers. It would be nice to have an integration test for it

PR #660, opened 2016-08-26
FUSETOOLS-2036: Project with Data Transformation cannot be run as Local Camel Context

jpav:


apupier:
This is not working for me.
In my TP, I have the 2.5.0 version and not the 2.4.3.
I checked in an installed version of JBDS 9.1 and it is a 2.5.0 version only.

lhein:
I don't think that this PR is needed at all. I updated the Jira issue with my findings. The project is not configured correctly and it still uses an old Camel facet which no longer exists.
Please try again with the project I attached to the Jira issue. If that works there is no reason to do this PR. If I missed the problem, please let me know.

lhein:
This PR doesn't fix the problem. The cause for it is a misconfiguration of the example project used in the Jira. Once that is configured properly it launches fine locally.
-1

PR #665, opened 2016-08-31
FUSETOOLS-2039: Cannot deploy a project with data transformation to JBoss Fuse 6.3.0

jpav:
Not sure if I'm testing this correctly, so can someone else let me know if they also see success?


lhein:
I created a new blank spring project. Then I added a timer and a log and a data transformation endpoint (XML -> XML) to the diagram. The update to the pom.xml was done and the config to the maven-bundle-plugin was added but it failed to build.

[INFO] --- maven-bundle-plugin:2.3.7:bundle (default-bundle) @ camel-spring ---
[ERROR] Bundle com.mycompany:camel-spring:bundle:1.0.0-SNAPSHOT : Unresolved references to [javax.xml.bind.annotation] by class(es) on the Bundle-Classpath[Jar:dot]: [org/apache/maven/pom/_4_0_1472621376291/Project.class, org/apache/maven/pom/_4_0_1472621377112/ObjectFactory.class, org/apache/maven/pom/_4_0_1472621377112/Project$Build$Plugins$Plugin.class, org/apache/maven/pom/_4_0_1472621376291/Project$Dependencies.class, org/apache/maven/pom/_4_0_1472621377112/Project$Repositories$Repository.class, org/apache/maven/pom/_4_0_1472621377112/package-info.class, org/apache/maven/pom/_4_0_1472621377112/Project$Build.class, org/apache/maven/pom/_4_0_1472621376291/Project$DependencyManagement$Dependencies.class, org/apache/maven/pom/_4_0_1472621376291/Project$DependencyManagement$Dependencies$Dependency.class, org/apache/maven/pom/_4_0_1472621377112/Project$DependencyManagement$Dependencies$Dependency.class, org/apache/maven/pom/_4_0_1472621376291/Project$Build$Plugins$Plugin.class, org/apache/maven/pom/_4_0_1472621376291/Project$PluginRepositories$PluginRepository$Releases.class, org/apache/maven/pom/_4_0_1472621376291/package-info.class, org/apache/maven/pom/_4_0_1472621377112/Project$Licenses.class, org/apache/maven/pom/_4_0_1472621376291/ObjectFactory.class, org/apache/maven/pom/_4_0_1472621376291/Project$Repositories$Repository.class, org/apache/maven/pom/_4_0_1472621377112/Project$Properties.class, org/apache/maven/pom/_4_0_1472621377112/Project.class, org/apache/maven/pom/_4_0_1472621376291/Project$Build$Plugins$Plugin$Configuration.class, org/apache/maven/pom/_4_0_1472621376291/Project$PluginRepositories$PluginRepository.class, org/apache/maven/pom/_4_0_1472621376291/Project$Properties.class, org/apache/maven/pom/_4_0_1472621377112/Project$Licenses$License.class, org/apache/maven/pom/_4_0_1472621377112/Project$Repositories$Repository$Releases.class, org/apache/maven/pom/_4_0_1472621376291/Project$PluginRepositories.class, org/apache/maven/pom/_4_0_1472621376291/Project$DependencyManagement.class, org/apache/maven/pom/_4_0_1472621376291/Project$PluginRepositories$PluginRepository$Snapshots.class, org/apache/maven/pom/_4_0_1472621377112/Project$Build$Plugins.class, org/apache/maven/pom/_4_0_1472621377112/Project$Repositories.class, org/apache/maven/pom/_4_0_1472621376291/Project$Build$Plugins.class, org/apache/maven/pom/_4_0_1472621377112/Project$Dependencies$Dependency.class, org/apache/maven/pom/_4_0_1472621376291/Project$Repositories$Repository$Snapshots.class, org/apache/maven/pom/_4_0_1472621377112/Project$DependencyManagement.class, org/apache/maven/pom/_4_0_1472621377112/Project$PluginRepositories.class, org/apache/maven/pom/_4_0_1472621377112/Project$PluginRepositories$PluginRepository.class, org/apache/maven/pom/_4_0_1472621377112/Project$PluginRepositories$PluginRepository$Releases.class, org/apache/maven/pom/_4_0_1472621377112/Project$Build$Plugins$Plugin$Configuration.class, org/apache/maven/pom/_4_0_1472621376291/Project$Licenses.class, org/apache/maven/pom/_4_0_1472621377112/Project$Dependencies.class, org/apache/maven/pom/_4_0_1472621377112/Project$DependencyManagement$Dependencies.class, org/apache/maven/pom/_4_0_1472621376291/Project$Repositories$Repository$Releases.class, org/apache/maven/pom/_4_0_1472621376291/Project$Dependencies$Dependency.class, org/apache/maven/pom/_4_0_1472621376291/Project$Licenses$License.class, org/apache/maven/pom/_4_0_1472621376291/Project$Build.class, org/apache/maven/pom/_4_0_1472621377112/Project$Repositories$Repository$Snapshots.class, org/apache/maven/pom/_4_0_1472621376291/Project$Repositories.class, org/apache/maven/pom/_4_0_1472621377112/Project$PluginRepositories$PluginRepository$Snapshots.class]
[ERROR] Error(s) found in bundle configuration
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE

So I wasn't even able to test the deploy to Fuse Runtime.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
```diff
@@ -361,17 +370,74 @@ public void init(IEditorSite site,
...
+	                                                                    "    <archive>" + //$NON-NLS-1$
+	                                                                    "        <manifestEntries>" + //$NON-NLS-1$
+	                                                                    "            <Project-Group-Id>${project.groupId}</Project-Group-Id>" + //$NON-NLS-1$
+                                                                        "            <Project-Artifact-Id>${project.artifactId}</Project-Artifact-Id>" + //$NON-NLS-1$
+                                                                        "            <Project-Version>${project.version}</Project-Version>" + //$NON-NLS-1$
+                                                                        "        </manifestEntries>" + //$NON-NLS-1$
+                                                                        "    </archive>" + //$NON-NLS-1$
+	                                                                    "</configuration>").getBytes()), //$NON-NLS-1$
+	                                          "UTF-8"); //$NON-NLS-1$
```
apupier: use StandardCharsets.UTF_8

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
```diff
@@ -361,17 +370,74 @@ public void init(IEditorSite site,
...
+	        IPath pomPath = project.getRawLocation() != null
+	                        ? project.getRawLocation()
+	                        : ResourcesPlugin.getWorkspace().getRoot().getLocation().append(project.getFullPath());
+	        Model pomModel = MavenPlugin.getMaven().readModel(new File(pomPath.append("pom.xml").toOSString()));  //$NON-NLS-1$
+	        Build build = pomModel.getBuild();
+	        Map<String, Plugin> pluginsByName = build.getPluginsAsMap();
+	        Plugin plugin = pluginsByName.get("org.apache.felix:maven-bundle-plugin"); //$NON-NLS-1$
+	        if (plugin == null) {
+	            plugin = new Plugin();
+	            plugin.setGroupId("org.apache.felix"); //$NON-NLS-1$
```
apupier: indentation is not the same everywhere

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
```diff
@@ -361,17 +370,74 @@ public void init(IEditorSite site,
...
+                plugin.setArtifactId("maven-bundle-plugin"); //$NON-NLS-1$
+                plugin.setVersion("2.3.7"); //$NON-NLS-1$
+                plugin.setExtensions(true);
+	            build.addPlugin(plugin);
+	        }
+	        Xpp3Dom config = (Xpp3Dom)plugin.getConfiguration();
```
apupier: shouldn't we check type before casting?

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
```diff
@@ -361,17 +370,74 @@ public void init(IEditorSite site,
...
+	                        : ResourcesPlugin.getWorkspace().getRoot().getLocation().append(project.getFullPath());
+	        Model pomModel = MavenPlugin.getMaven().readModel(new File(pomPath.append("pom.xml").toOSString()));  //$NON-NLS-1$
+	        Build build = pomModel.getBuild();
+	        Map<String, Plugin> pluginsByName = build.getPluginsAsMap();
+	        Plugin plugin = pluginsByName.get("org.apache.felix:maven-bundle-plugin"); //$NON-NLS-1$
+	        if (plugin == null) {
+	            plugin = new Plugin();
+	            plugin.setGroupId("org.apache.felix"); //$NON-NLS-1$
+                plugin.setArtifactId("maven-bundle-plugin"); //$NON-NLS-1$
+                plugin.setVersion("2.3.7"); //$NON-NLS-1$
```
apupier: the current latest version is 3.2.0

lhein:
It is needed to list both the * and the com.sun.el;version="[2,3)";resolution:=optional inside the Import-Package statement. Having only one or the other results in an error either on build time or on deploy time.

Can you please also provide a test for this change?

apupier:
As we are putting it in MANIFEST.MF, how will it work for EAP when using a war?

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
```diff
@@ -361,17 +370,74 @@ public void init(IEditorSite site,
             if (!latestVersion) prefs.setValue(VERSION_PREFERENCE, version);
 
             // Ensure Maven will compile transformations folder
-            final IProject project = manager.project();
+            IProject project = manager.project();
```
apupier: The method TransformationEditor.init is 100 lines long. We shoudl split it for better readability.
For instance I see several distinct part (which can surely subdivided in soelm cases):
- create Transformation Folder
- ensure Pom contains MANIFEST.MF configuration
- compile project

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
```diff
@@ -361,17 +370,74 @@ public void init(IEditorSite site,
...
+	        if (!importPkgs.contains("com.sun.el;version=")) { //$NON-NLS-1$
+	            if (!importPkgs.isEmpty()) importPkgs += ",\n"; //$NON-NLS-1$
+	            importPkgs += "*,com.sun.el;version=\"[2,3)\";resolution:=optional"; //$NON-NLS-1$
+	            importPkg.setValue(importPkgs);
+	            try (OutputStream out = new BufferedOutputStream(new FileOutputStream(pomFile))) {
+	                MavenPlugin.getMaven().writeModel(pomModel, out);
+	            }
+	        }
+
             IProgressMonitor monitor = new NullProgressMonitor();
```
apupier: For another PR: better to use a ProgressMonitorDialog or at least a busyCursor

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
jpav: Not sure what's going on with the indentation.  It looks fine locally, and I've checked my prefs, and they're set to insert spaces for tabs.

apupier: I would prefer to let the default Eclipse configuration which is using tab

image

Maybe youcan  try Ctrl+A, Ctrl+I to reformat the whole file.

lhein: Thats probably the point. I don't use spaces for tabs.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
jpav: I'll add a better message, but the editor will end up failing anyways if this cast doesn't work.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
jpav: I'll separate out this change for a test, but as an aside...

I know many people talk about having separate methods for readability, but in all my experience, the opposite is achieved by separating the logic into lots of smaller methods, which appear who knows where within the code, just for readability and not reusability.  However, it does very much lend itself to testability.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
jpav: Well, crap.  I didn't realize we were different on that.  In my experience, this can become problematic in situations where code has to be changed in environments where an Eclipse editor either isn't used or isn't available.  Obviously this doesn't apply to the UI stuff, but more backend code.  I'll go ahead and change my settings though to avoid my settings battling everyone else's.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
jpav: Regarding the WAR comment, I have no idea.  @lhein?

apupier:
for the WAR, I think that we need to test it and then ask to Fuse Runtime and Fuse on EAP guys

apupier:
it seems that it misses the parent pom.xml update to include the new transformation.editor.tests.integration module

jpav:
Crap, I forgot about that…

transformation/tests/org.jboss.tools.fuse.transformation.editor.tests.integration/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditorIT.java
```diff
@@ -0,0 +1,176 @@
+/******************************************************************************
+ * Copyright (c) 2015 Red Hat, Inc. and others.
```
apupier: 2016

transformation/tests/org.jboss.tools.fuse.transformation.editor.tests.integration/src/main/java/org/jboss/tools/fuse/transformation/editor/FuseProject.java
```diff
@@ -0,0 +1,56 @@
...
+ *****************************************************************************/
+package org.jboss.tools.fuse.transformation.editor;
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.NullProgressMonitor;
+import org.junit.rules.ExternalResource;
+
+public class FuseProject extends ExternalResource {
```
apupier: why not simply reusing org.fusesource.ide.camel.model.service.core.tests.integration.core.io.FuseProject ?

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
```diff
@@ -514,4 +529,62 @@ void updateHelpText() {
...
+        Plugin plugin = pluginsByName.get("org.apache.felix:maven-bundle-plugin"); //$NON-NLS-1$
+        if (plugin == null) {
+            plugin = new Plugin();
+            plugin.setGroupId("org.apache.felix"); //$NON-NLS-1$
+            plugin.setArtifactId("maven-bundle-plugin"); //$NON-NLS-1$
+            plugin.setVersion("3.2.0"); //$NON-NLS-1$
+            plugin.setExtensions(true);
+            build.addPlugin(plugin);
+        }
+        if (!(plugin.getConfiguration() instanceof Xpp3Dom)) throw new Exception(Messages.TransformationEditor_invalidPomConfiguration);
```
lhein: This line will be a problem if the plugin was not defined in the pom before OR if it had no configuration. In that case the instanceof call will return false and you throw a new Exception here which makes it impossible then to display the transformation editor at all.

lhein: Another side effect is that the new plugin is not persisted into the pom if the above occurs.

lhein:
Deploy to EAP seems to work. At least I wasn't able to spot any exceptions.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
lhein: created a PR on your fork with a draft how to fit it:
https://github.com/jpav/fuseide/pull/1

transformation/tests/org.jboss.tools.fuse.transformation.editor.tests.integration/src/main/java/org/jboss/tools/fuse/transformation/editor/FuseProject.java
jpav: I thought about that, but I was trying to avoid what seemed like an odd dependency to another test project that had nothing to do with transformations.  Plus, when I first created the test, I was planning on using multiple instances of FuseProject, each with different POM contents, and the auto-setting of those contents in your class's constructor shied me away from that.  I guess there's no great reason not to use it though, and would avoid a mostly duplicate class.  No one has problems with the weird dependency?  If not, I'll change this.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
jpav: So, I originally added this to address @apupier earlier comment about an unchecked type cast, but after researching this more, it appears that the only possible instance of the configuration will be a Xpp3Dom, so the original type cast should have been okay.  The suggestion you provided in the fork, @lhein, allows for an "additional" configuration if the instance isn't what's expected, and since that's not possible, I'm going to revert back to the blind cast.  For what it's worth, all examples of working with the configuration that I can find online do the same thing.

transformation/tests/org.jboss.tools.fuse.transformation.editor.tests.integration/.classpath
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
+	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
+	<classpathentry kind="src" path="src/main/java"/>
+	<classpathentry kind="src" path="src/main/resources"/>
```
apupier: this folder doesn't exist

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
```diff
@@ -361,17 +371,22 @@ public void init(IEditorSite site,
...
-            final IProject project = manager.project();
+            IProject project = manager.project();
 			File pomFile = project.getLocation().append("pom.xml").toFile(); //$NON-NLS-1$
-			final MavenUtils mavenUtils = new MavenUtils();
+			MavenUtils mavenUtils = new MavenUtils();
 			mavenUtils.addResourceFolder(project, pomFile, Util.TRANSFORMATIONS_FOLDER);
 			mavenUtils.addResourceFolder(project, pomFile, MavenUtils.RESOURCES_PATH);
 
+			// Ensure import of package com.sun.el (FUSETOOLS-2039)
+			updateManifestPackageImports(project, pomFile);
```
apupier: I think that the maven-bundle-plugin should be added only if the packaging type is of type bundle (or perhaps jar?) It has nothing to do for a war project

apupier:
to fix the build, the package org.jboss.tools.fuse.transformation.editor needs to be exported from org.jboss.tools.fuse.transformation.editor plugin (preferably with an x-friends to org.jboss.tools.fuse.transformation.editor.tests.integration

transformation/tests/org.jboss.tools.fuse.transformation.editor.tests.integration/src/main/java/org/jboss/tools/fuse/transformation/editor/FuseProject.java
apupier: if there is a weird dependency we can extract the the targeted class in a "common.test" or "common.util.test" plugin

transformation/tests/org.jboss.tools.fuse.transformation.editor.tests.integration/build.properties
```diff
@@ -0,0 +1,5 @@
+source.. = src/main/java/,\
+           src/main/resources/
```
apupier: src/main/resources doesn't exist

jpav: Ah, this is due to github not creating empty folders.  I have it locally.  I'll delete it.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
jpav: k, I'll add a check then to skip all this if it's a war.

transformation/tests/org.jboss.tools.fuse.transformation.editor.tests.integration/build.properties
apupier: yes this is a boring thing of git (and not github)

transformation/tests/org.jboss.tools.fuse.transformation.editor.tests.integration/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditorIT.java
```diff
@@ -0,0 +1,193 @@
...
+			"  <modelVersion>4.0.0</modelVersion>\n" +
+			"  <groupId>com.mycompany</groupId>\n" +
+			"  <artifactId>testproject</artifactId>\n" +
+			"  <version>1.0.0-SNAPSHOT</version>\n" +
+			"  <packaging>bundle</packaging>\n" +
+			"  <name>Some Dummy Project</name>\n" +
+			"  <build>\n" +
+			"    <plugins>\n";
+
+	private static final String POM_END = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
```
apupier: first big issue, this line should not exists
"<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"

transformation/tests/org.jboss.tools.fuse.transformation.editor.tests.integration/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditorIT.java
```diff
@@ -0,0 +1,193 @@
...
+
+	private void updatePom(String pom) throws Exception {
+		updatePom(pom, EXPECTED_POM);
+	}
+
+	private void updatePom(String pom,
+	                       String expectedPom) throws Exception {
+		pomIFile.create(new ByteArrayInputStream(pom.getBytes()), true, monitor);
+		editor.updateManifestPackageImports(project, pomFile);
+		char[] buf = new char[expectedPom.length()];
```
apupier: another issue is here: we need to use the size of the inputstream not the expected pom because the size can be different due to space/tabs/empty lines

apupier:
Does the <excludeDependencies>false</excludeDependencies> required? If yes, this use case is not covered if there is a configuration provided without specifying it

apupier:
see https://github.com/jpav/fuseide/pull/2 for a fix on the currently failing tests

apupier:
After workign on it a bit:
- Shouldn't the workaround provided only when the version targeted is a Fuse Runtime 6.2.x ? No need for it in 6.3, right?
- I'm ok to handle it like that as a workaround for the coming release, but I think that the tooling should generate stuff for the right target. Here we are modifying the Runtime due to a Tooling limitations, it seems really bad

apupier:
another question: are we supposed to run on JVm other than the hotspot one?
I suppose that com.sun.el is not part of IBM JVM fo rinstance

lhein:
In JBoss Fuse 6.3 this package is exported by the bundle with symbolic name: org.glassfish.web.javax.el

This means its not coming from the Java Runtime and therefore it shouldn't be an issue.

lhein:
> Shouldn't the workaround provided only when the version targeted is a Fuse Runtime 6.2.x ? No need for it in 6.3, right?

--> if it makes problems on 6.2.x then we should limit it to 6.3....if not then I see no reason to introduce a version range specific behaviour without a dire need.

> I'm ok to handle it like that as a workaround for the coming release, but I think that the tooling should generate stuff for the right target. Here we are modifying the Runtime due to a Tooling limitations, it seems really bad

--> I do not agree on this one. We do not modify the runtime. We just improve the package imports of the bundle because the maven-bundle-plugin doesn't do it for us. Why has the tooling a limitation here?

apupier:
> Shouldn't the workaround provided only when the version targeted is a Fuse Runtime 6.2.x ? No need for it in 6.3, right?
> --> if it makes problems on 6.2.x then we should limit it to 6.3....if not then I see no reason to introduce a version range specific behaviour without a dire need.

the dire need is simplicity, why introducing an export package if not required? It might cause some integration issues

> 
> I'm ok to handle it like that as a workaround for the coming release, but I think that the tooling should generate stuff for the right target. Here we are modifying the Runtime due to a Tooling limitations, it seems really bad
> --> I do not agree on this one. We do not modify the runtime. We just improve the package imports of the bundle because the maven-bundle-plugin doesn't do it for us. Why has the tooling a limitation here?

If I understood well the issue, the Tooling genrates the jackson with EL in the "wrong" version, not the same than the runtime. If the tooling is able to generate with the "right" version -i.e. with the same than the Runtime - the export package is useless.

lhein:
Can you point me to the discussion where this was exposed? When looking at the original issue it is a conflict of 2 bundles exposing the same package in the runtime. Gregorz eliminated one of the bundles and that caused an issue. 
When looking at the comments of ENTESB-5411 it seems to be solely an OSGi resolving problem of the camel-dozer component.

apupier:
In discussion on Fuse Eclipse Tooling HipChat chat room, it was the status done which I summarized in this comment: https://issues.jboss.org/browse/FUSETOOLS-2039?focusedCommentId=13284724&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13284724

I can't believe that the Runtime is now requiring to export this package for all the Route packages, this is a big constraints.

apupier:
Ok after discussions with @lhein seems that I mixed up import and export :-)

So my understanding of the issue is:
- before we had a direct dependency which was importing the javax.el
- now it is no more the case so we need to add the import package on con.sun.el which is provided by Fuse Runtime

lhein:
I agree. It must be like you wrote.

apupier:
as @jpav stopped working on this issue, I created my own branch and created another PR to continue on what has been done already here https://github.com/fusesource/fuseide/pull/671

PR #666, opened 2016-08-31
FUSETOOLS-2071 - Set Fuse bom version when Target runtime selected

apupier:
- currently rely on the fact the UI will have restrained the list of
possible choice and that we have a 1-1 mapping between fuse bom version
and camel version

### DIFF
--- a/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/configurators/MavenTemplateConfigurator.java
+++ b/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/configurators/MavenTemplateConfigurator.java
@@ -95,7 +95,7 @@ public class MavenTemplateConfigurator extends DefaultTemplateConfigurator {
 	}
 	
 	/**
-	 * changes all occurances of Camel version in the pom.xml file with the
+	 * changes all occurrences of Camel version in the pom.xml file with the
 	 * version defined in the wizard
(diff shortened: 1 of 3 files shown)


editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/configurators/MavenTemplateConfigurator.java
```diff
@@ -124,6 +124,10 @@ protected boolean configurePomCamelVersion(IProject project, NewProjectMetaData
 			
 			if(projectMetaData.getTargetRuntime() == null){
 				MavenUtils.alignFuseRuntimeVersion(m2m, camelVersion);
+			} else {
+				// we suppose that only one version of Fuse Runtime is possible for a Camel Version
+				//TODO: find a way to retrieve the Fuse Runtime BOM version from the Target Runtime
+				MavenUtils.alignFuseRuntimeVersion(m2m, camelVersion);
 			}
```
lhein: couldn't we just always call the align and skip the if / else?

apupier: I prefer to let it clear in the code that we rely on an assumption which is not perfect

lhein:
ok, +1 then

lhein:
do we have a test in place?

PR #668, opened 2016-09-01
FUSETOOLS-2051 - fix disabled Finish button issue

lhein:


apupier:
might worth adding few assertions in FuseIntegrationProjectWizardTemplatePageIT

lhein:
That basically not possible in the current state as we don't go through each wizard page. So basically page 1 and 2 of the wizard are not filled yet and the wizard will always return false for canFinish. No idea what else u had in mind.

lhein:
@apupier No clue why the eventing is broken in that test. Basically none of the widget listeners are being called when changing button selections etc. Not sure if that is normal. I had to emulate the behaviour done by the listeners which makes the test itself a bit useless imo. Having a real UI test would probably be better in this case. Thoughts?

apupier:
No idea why the event mechanism is not working in the test.
We can ask QE to cover this little steps in their UI tests (I suppose that they already have a test for this wizard)
+1 then

lhein:
testPR

PR #670, opened 2016-09-01
JBTIS-844 - update fuse tooling to 4.3.3.Final and tycho to 0.26

pleacu:
@lhein @apupier - this works for me - cleaned out my local m2 repo and build fuse tooling and the fuse TP

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,11 +41,11 @@
     <maven.compiler.target>1.8</maven.compiler.target>
 
     <!-- JBoss Tools Integration Stack version -->
-    <jbtis.version>4.3.1.Final</jbtis.version>
+    <jbtis.version>4.3.3.Final</jbtis.version>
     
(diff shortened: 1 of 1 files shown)


lhein:
It seems Jenkins has a different opinion :)

apupier:
testPR

apupier:
on which maven repo is supposed to be the org.jboss.tools.integration-stack:target-platform:target:base-ea:4.3.3.Final ?
@pleacu maybe you have a specific repo in your .settings which is missing in our pom.xml

apupier:
so after i relaunched it, it works, so maybe the repo was down the first time.

apupier:
I merge this PR in order to go further with EAP tests locally.
If someone want to update the TP locally, he needs to use the 4.3.4-SNAPSHOT version for now. @pleacu will provide a 4.3.4.Final soon I suppose.

PR #673, opened 2016-09-07
FUSETOOLS-2085 - automate build of the project templates and fixed so…

lhein:
…me errors in template readmes and maven config


editor/plugins/org.fusesource.ide.projecttemplates/.descriptors/blank-spring.xml
```diff
@@ -0,0 +1,17 @@
...
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.3 http://maven.apache.org/xsd/assembly-1.1.3.xsd">
+	<id>blank-spring</id>
+	<includeBaseDirectory>false</includeBaseDirectory>
+	<formats>
+		<format>zip</format>
+	</formats>
+	<fileSets>
+		<fileSet>
+			<directory>${basedir}/../../../templates/blank/blank-spring</directory>
```
apupier: what do you think of moving the templates folders inside the org.fusesource.ide.projecttemplates plugin, it will avoid to have such relative paths and it will help to gather related things together.
We will just need to exclude the folder in build.properties

lhein: We can do that but keep in mind that the template sources will be hidden in a 4th level subfolder instead of being a direct subfolder in the root of the repo. Thats probably not a problem for us but could confuse anyone else.

apupier: confuse who?
the template are not meant to be imported from other way than Fuse Tooling.
It seems easier to find the tempaltes in the plugin which is providing them.

lhein: anyone who wants to look at those sources or contribute new ones. but I am fine with that change. We can document the procedure here at github.

bfitzpat:
That's pretty awesome. I was hoping we'd get to the point where the build of those zips would become automated. +1

lhein:
testPR

lhein:
testPR

lhein:
testPR

editor/plugins/org.fusesource.ide.projecttemplates/pom.xml
```diff
@@ -1,16 +1,67 @@
...
+						<descriptor>.descriptors/simple-amq-spring.xml</descriptor>
+						<descriptor>.descriptors/medium-eap-wildfly-spring.xml</descriptor>
+						<descriptor>.descriptors/medium-cxf-codefirst-java.xml</descriptor>
+						<descriptor>.descriptors/medium-cxf-codefirst-spring.xml</descriptor>
+					</descriptors>
+					<outputDirectory>templates</outputDirectory>
+				</configuration>
+				<executions>
+					<execution>
+						<phase>compile</phase>
```
apupier: shouldn't we bind it to generate-resources which is dedicated to "generate resources for inclusion in the package.?
https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference

(even if i'm not sure that it makes a big differenc ein our case, perhaps just that we migth be able to call generate-resources phase only to go faster to regenerate the zip. which lead me think, what about a dedicated launch configuration for that?)

apupier: see my PR implementing my feedback:https://github.com/lhein/fuseide/pull/32

apupier:
we also need to add a .gitignore in templates folder for zip files to avoid polluting the git stating area and commit the zipped templates by error

templates/simple/amq/simple-fuse-activemq-blueprint/src/main/java/com/mycompany/templates/camel/activemq/OrderGenerator.java
```diff
@@ -1,45 +0,0 @@
-/**
```
apupier: Why did you removed this file? it is due to that the test is failing. The project is not compiling due to this error

templates/simple/amq/simple-fuse-activemq-spring/src/main/java/com/mycompany/templates/camel/activemq/OrderGenerator.java
```diff
@@ -1,45 +0,0 @@
-/**
```
apupier: Why did you removed this file? it is due to that the test is failing. The project is not compiling due to this error

lhein:
testPR

.gitignore
```diff
@@ -12,3 +12,5 @@ runtime-EclipseApplication-m2eclipse
 ./fuseenterprise
 .idea
 libs/
+
```
lhein: Yeah, but I always put an empty line at the bottom. Its not really crucial to remove this line I'd say.

apupier:
+1 (better if empty lines are removed from .gitignore but not mandatory)

.gitignore
apupier: there are 2 empty lines ^^

but yes not a mandatory change

PR #685, opened 2016-09-14
FUSETOOLS-2060 - Always return non-null value for iconName

apupier:
there is another issue which is that the referenced endpoint is not resolved, but we will have the default icon instead of a NPE. Seems to be a first good step.

### DIFF
--- a/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
+++ b/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
@@ -1535,26 +1535,24 @@ public abstract class AbstractCamelModelElement {
 		if (isEndpointElement()) {
 			String u = (String) getParameter(URI_PARAMETER_KEY);
 			if (u != null && u.trim().length() > 0) {
-				String scheme = null;
 				if (u.startsWith("ref:")) {
 					// if its a ref we lookup what is the reference scheme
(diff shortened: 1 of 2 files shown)


core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
```diff
@@ -1535,26 +1535,24 @@ public String getIconName() {
...
 				if (u.startsWith("ref:")) {
 					// if its a ref we lookup what is the reference scheme
-					String refId = u.substring(u.indexOf(":") + 1);
+					String refId = u.substring(u.indexOf(':') + 1);
 					AbstractCamelModelElement endpointRef = getCamelContext().getEndpointDefinitions().get(refId);
 					if (endpointRef != null) {
 						String refUri = (String) endpointRef.getParameter(URI_PARAMETER_KEY);
 						if (refUri != null) {
-							scheme = refUri.substring(0, refUri.indexOf(":"));
+							return refUri.substring(0, refUri.indexOf(':'));
```
lhein: should we cover the case where no : is in the uri for whatsoever reason?

lhein:
+1 on success

apupier:
testPR

apupier:
previous job build failed, I investigated a bit but cannot find something useful to avoid the isseu again. i think that it is not related to this PR and an "asynchrone" issue. The only stacktraces that I found suspect were already reported on Eclipse bugtracker.

PR #687, opened 2016-09-15
FUSETOOLS-2102 - Recommit zipped templates

apupier:
- using assembly was overriding the main jar artifact during maven
install phase
- kept a property to reactivate it in order to easily regenerate zipped
templates during development


lhein:
+1

PR #688, opened 2016-09-15
FUSETOOLS-2102 - replacing assembly plugin with truezip plugin

lhein:


apupier:
we need to remove the currently commited zipped templates (and/or readd the .gitignore )

apupier:
Why do we have skip set to true for maven-clean-plugin?
it was here before but shouldn't we remove it now?

lhein:
we can remove the skip I guess

lhein:
ok, still failing tests. will look at it next week.

lhein:
testPR

apupier:
testPR

apupier:
+1

PR #689, opened 2016-09-16
FUSETOOLS-2039 - Handle Export Package with Data Transformation

apupier:
- in case of using a commited MANIFEST.MF: complete the export-package
list in the Manifest.mf file
- in case of using maven-bundle-plugin:
-- if no export package configured at all: do nothing
-- if some export package configured: complete the export-package list
in the pom.xml for the maven-bundle-plugin

### DIFF
--- a/transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/ImportExportPackageUpdater.java
+++ b/transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/ImportExportPackageUpdater.java
@@ -17,10 +17,8 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.StandardCharsets;
-import java.util.HashMap;
+import java.util.Arrays;
 import java.util.Map;
-import java.util.jar.Attributes;
-import java.util.jar.Manifest;
 
 import org.apache.maven.model.Build;
 import org.apache.maven.model.Model;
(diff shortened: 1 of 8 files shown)


apupier:
I export also the package of Java class specified in case of "Java" transformation:
- if it is in the JDK --> this export might be strange but i'm not sure to be able to determine this use case
- otherwise only classes available in the dependencies is available, so it is a reexport, seems to be the right way to do as this bundle requires to export this package in order that the consumer use the described transformation.
In case i'm wrong, we might export only if the package comes from the bundle itself in the Java case.
Thought?
For Another PR perhaps?

lhein:
I would think it's enough to export packages of the bundle.  Dependencies should be bundles on the runtime imo.

apupier:
> I would think it's enough to export packages of the bundle. Dependencies should be bundles on the runtime imo.

and for another PR or required for this one?
I started to have a look how to it.

lhein:
I would keep it here.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/util/ImportExportPackageUpdater.java
```diff
@@ -199,7 +269,7 @@ private String addELPackage(String importPkgs) {
 			importPkgs = "";
 		}
 		if (!importPkgs.isEmpty()){
-			importPkgs += ",\n"; //$NON-NLS-1$
+			importPkgs += ",\n "; //$NON-NLS-1$
```
lhein: should we care about max line widt / max chars per statement in the manifest and do the appropriate line break if needed?

apupier: - will we have one package which will be longer than the max line width?
- During Jar generation the manifest is not reformatted? (it is the case for Tycho but i'm not familiar with other way of doing it)

apupier: yes, package length can be longer (and classname is not part of it)
Manifest files requires to be 72-characters long but they support 512 in fact.
I rely on Eclipse API to write the Manifest file. They are not stripping the content for readability purpose and explains that it should be done during compilation, see https://bugs.eclipse.org/bugs/show_bug.cgi?id=91309
So I would prefer to avoid dealing with this kind of thing in our code.

lhein: ok, agreed

lhein: +1

apupier:
testPR

apupier:
relaunch, as another PR was triggered just after I can't investigate enough on the issue

apupier:
testPR

PR #695, opened 2016-09-19
FUSETOOLS-2053 - fix problem with camel version preselection in wizard

lhein:


bfitzpat:
+1

lhein:
Its a fairly small change. The only thing I do here is to filter camel versions which don't fit the runtimes camel version in terms of branded vs community camel versions. It means if the runtime contains a Red Hat branded Camel version (it always will) then we only suggest Red Hat branded versions for that runtime and NOT like it is now 2.17.3 community version.

apupier:
ok, I still think that a unit test will help to understand that

editor/tests/org.fusesource.ide.projecttemplates.tests.integration/src/main/java/org/fusesource/ide/projecttemplates/wizards/pages/FuseIntegrationWizardRuntimeAndCamelPageIT.java
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * @author lhein
+ *
+ */
+public class FuseIntegrationWizardRuntimeAndCamelPageIT {
+	
+	@Test
+	public void testCompatibleCamelVersionSelectedForUnsupportedBrandedVersion() throws Exception {
+		FuseIntegrationProjectWizard wiz = new FuseIntegrationProjectWizard();
```
apupier: What about extracting these four lines in a setup @before method?

lhein:
testPR

lhein:
testPR

PR #699, opened 2016-09-22
FUSETOOLS-2061 - Respect Namespace for Global Definitions

apupier:
- avoids to put ids for elements whcih don't support it
- avoids to show unsupported and unknown elements in Global
Configuration Editor that are part of other namespaces

Requires a validation from a Camel point of view:
- Am I missing some cases?
- Are we ok to hide elements in Configuration Editor that are not part of camel namespaces? (they can't be modified anyways)


bfitzpat:
Sorry but I don't know enough about this chunk of code to approve/deny...

apupier:
Discussion summary:
- not OK to hide not supported elements
- need to add ids only on supported elements
- need to handle global configuration elements without ids

apupier:
testPR

apupier:
with updated PR:
- show elements unknown without ids as <unknownId> (<tagName>)
- add ids only on known elements/namespaces
- cannot delete through UI an unknown elements (need to use source tab)

image

PR #710, opened 2016-09-28
code improvement for deployment which did not make it into 8.0

lhein:


lhein:
@apupier pls review

apupier:
@bfitzpat from the javadoc on the new method:

> Returns IMavenProjectFacade corresponding to the pom. This method first looks in the project cache, then attempts to load the pom if the pom is not found in the cache. In the latter case, workspace resolution is assumed to be enabled for the pom but the pom will not be added to the cache.

so I think we are better here

PR #711, opened 2016-09-29
FUSETOOLS-2137 - Add missing bundle-vendor header

apupier:


apupier:
"can we have Bundle-Vendor: %Bundle-Vendor instead ?"
Do we want to translate the Vendor name? I supposed that we want to keep it always to Red hat

lhein:
I was just comparing with what we use elsewhere in the codebase. And it seems we used localization everywhere.

lhein:
looks fine except the merge conflicts

PR #717, opened 2016-10-06
FUSETOOLS-2144 - ensuring that the list of supported models is

bfitzpat:
initialized before being iterated through.
- added code to ensure initialization
- added unit test to test the method, which wasn't being tested
previously

### DIFF
--- a/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactory.java
+++ b/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactory.java
@@ -181,6 +181,10 @@ public class CamelModelFactory {
 	 */
 	public static String getLatestCamelVersion() {
 		String latest = null;
+		// make sure we're initialized here
+		if (supportedCamelModels == null || supportedCamelModels.keySet() == null) {
+		    initializeModels();
+		}
 		for (String v : supportedCamelModels.keySet()) {
 			if (latest == null) {
 				latest = v;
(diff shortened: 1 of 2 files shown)


core/tests/org.fusesource.ide.camel.model.service.core.tests/src/test/java/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactoryTest.java
```diff
@@ -27,6 +27,12 @@
 																		 "2.15.1",
 																		 "2.17.0.redhat-630077", 
 																		 "2.17.0.redhat-630159");
+	@Test
+	public void testGetLatestCamelVersion() {
```
apupier: camel model factory are initialized by extension points, so this test needs to be an integration one.

core/tests/org.fusesource.ide.camel.model.service.core.tests/src/test/java/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactoryTest.java
```diff
@@ -27,6 +27,12 @@
...
+	    String version = CamelModelFactory.getLatestCamelVersion();
+	    assertThat(version).isNotNull();
```
apupier: no need to assert not null as we assert just after that it is equals to 2.17.3

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/CamelModelFactory.java
```diff
@@ -181,6 +181,10 @@ public static String getCompatibleCamelVersion(String requestedCamelVersion) {
 	 */
 	public static String getLatestCamelVersion() {
 		String latest = null;
+		// make sure we're initialized here
+		if (supportedCamelModels == null || supportedCamelModels.keySet() == null) {
+		    initializeModels();
```
apupier: if called from several places, I think that we need to create a singleton and have a synchronized block

bfitzpat:
I'm going to close this. Turns out it was an issue with the test on the SY side (not surprising) and not on the Fuse side.

PR #722, opened 2016-10-13
FUSETOOLS-1482 FUSETOOLS-2087 - Improve path parameter support

apupier:
- support path parameters starting with slashes
- support when there are several parameters


apupier:
testPR

PR #727, opened 2016-10-17
Fusetools 2161 demo the flow

apupier:


apupier:
was here for demo purpose

PR #728, opened 2016-10-17
FUSETOOLS-2161 - sysout added

apupier:


apupier:
was here for demo purpose

PR #730, opened 2016-10-18
FUSETOOLS-2131 - provide preference page to configure staging

apupier:
repositories

image

I use directly a String to specify the repositories, it is how they are stored in preferences.
A very custom Preference page would be required to have something easier to use.
As the main objective is to ease use for internal usage when we are based on a not yet released version, i think that it is reasonable. What do you think?


bfitzpat:
Any way we could turn the ugly comma-separated-list into a listview for easier consumption? Though this is a pretty advanced feature for our users, I would think we might want to make it easier to use?

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/l10n/messages.properties
```diff
@@ -66,4 +66,9 @@ newProjectWizardTemplatePageTemplateProjectLabel = Use a predefined template
 newProjectWizardTemplatePageTemplateProjectDescription = Select this option if you would like to start with one of our predefined project templates.
 newProjectWizardTemplatePageDSLLabel = Select the project type
 newProjectWizardTemplatePageFilterBoxText = type filter text (filters on name, description and keywords)
+StagingRepositoriesFieldEditor_ErrorMessageDuplicateRepoId=Two repositories have the same name: 
+StagingRepositoriesFieldEditor_ErrorMessageWrongPattern=The staging repositories need to follow the pattern <repositoryName1>,<repositoryURL1>;<repositoryName2>,<repositoryURL2>
```
bfitzpat: I would change this warning text slightly to- "Staging repositories must follow the pattern <repositoryName1>,<repositoryURL1>;<repositoryName2>,<repositoryURL2>" just to reduce wordiness a bit.

apupier:
> Any way we could turn the ugly comma-separated-list into a listview for easier consumption? Though this is a pretty advanced feature for our users, I would think we might want to make it easier to use?

see my previous comment:
I use directly a String to specify the repositories, it is how they are stored in preferences.
**A very custom Preference page would be required** to have something easier to use.
As the main objective is to ease use for internal usage when we are based on a not yet released version, i think that it is reasonable. What do you think?

bfitzpat:
Yes, that makes sense.

apupier:
@lhein does it make sense for you too?

or should we delay the creation of the preference page to 9.1 and take time to provide a better UI (I don't believe to have time to do it for 9.0)

lhein:
I see no immediate need having this for 9.0. I'd rather move that to 9.1.

lhein:
btw. the PR seems to have a wrong Jira issue number

lhein:
#758 seems to solve the same thing. I guess we may close one of these PRs.

apupier:
PR https://github.com/fusesource/fuseide/pull/758 is more complete, I close this one

PR #733, opened 2016-10-18
FUSETOOLS-2138 - be more robust against InterruptedException

apupier:
I hit InterruptedException 2 times (out of 4) during investigation so I improved robustness


transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
```diff
@@ -379,23 +379,25 @@ public void init(IEditorSite site,
...
-            waitJavaBuild(monitor);
+            waitJavaBuild(20, null, monitor);
         } catch (final Exception e) {
             throw new PartInitException("Error initializing editor", e); //$NON-NLS-1$
         }
     }
 
-	private void waitJavaBuild(IProgressMonitor monitor) {
+	private void waitJavaBuild(int decreasingCounter, InterruptedException interruptedException, IProgressMonitor monitor) {
+		if (decreasingCounter > 0) {
```
jpav: Shouldn't this be if (decreasingCounter == 0)?

PR #739, opened 2016-10-19
FUSETOOLS-2123 - fix for the flashing progress dialog

bfitzpat:


bfitzpat:
Updated PR - removed unnecessary import. Don't really have a solution to the other issue.

PR #740, opened 2016-10-20
FUSETOOLS-2157: Root element name must be supplied for the transformation.

jpav:


targetplatform/JBTIS Mars Mac OS.launch
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
apupier: I would prefer to have the .launch file in a separated PR

PR #742, opened 2016-10-20
FUSETOOLS-2166 - Avoid losing path parameters when going back and forth

apupier:
- it was occurring because the ModifyListener is calling an update on
the path parameters before the modelMap was completely initialized
- the solution consists to fall back to current model state if modelMap
is not initialized for the parameter required


apupier:
thanks for the fast review guys

PR #747, opened 2016-10-28
Clean code - use interface instead of Implementation

apupier:
this is cleaner to use the Interface and it will be required if we want to provide a Jolokia Implementation for debugging. this will allow to debug camel routes deployed on IoT devices using Eclipse Kura for instance


lhein:
I am not sure I see the benefit of doing this.

apupier:
"I am not sure I see the benefit of doing this."
--> always better to use Interfaces than concrete implementation

PR #754, opened 2016-11-08
FUSETOOLS-2170 - fixing missing dependency

bfitzpat:
I don't like this solution. I suspect the dependency isn't properly registered upstream from us somewhere, but this should take care of it at our level.

### DIFF
--- a/core/plugins/org.fusesource.ide.maven/META-INF/MANIFEST.MF
+++ b/core/plugins/org.fusesource.ide.maven/META-INF/MANIFEST.MF
@@ -21,5 +21,6 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.m2e.refactoring,
  org.eclipse.m2e.scm,
  org.eclipse.m2e.maven.runtime;bundle-version="1.6.0",
- org.eclipse.m2e.maven.runtime.slf4j.simple;bundle-version="1.6.0"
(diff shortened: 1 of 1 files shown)


core/plugins/org.fusesource.ide.maven/META-INF/MANIFEST.MF
```diff
@@ -21,5 +21,6 @@ Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.m2e.refactoring,
  org.eclipse.m2e.scm,
  org.eclipse.m2e.maven.runtime;bundle-version="1.6.0",
- org.eclipse.m2e.maven.runtime.slf4j.simple;bundle-version="1.6.0"
+ org.eclipse.m2e.maven.runtime.slf4j.simple;bundle-version="1.6.0",
+ org.sonatype.tycho.m2e;bundle-version="0.9.0";visibility:=reexport
```
apupier: I think that the visibility should not be reexport. If another plugin require sit it needs to include it in required dependency

lhein: yeah, please align with Aureliens PR.

apupier:
I created https://github.com/fusesource/fuseide/pull/756 for the fix on the master

bfitzpat:
Thanks guys. I'll go ahead and close this PR since you have it taken care of elsewhere.

PR #762, opened 2016-11-16
Normalize all the line endings

apupier:
use Linux style

need to check that we have the correct result with this PR


apupier:
I created https://github.com/apupier/fuseide/pull/3 to try to check for line endings but seems that it is not doing the trick

lhein:
can you check your ~/.gitconfig file?

mine contains:

[core]
      	autocrlf = false
        safecrlf = true

apupier:
I have (as I configured previously)
[core]
	autocrlf = true

lhein:
please try my settings

apupier:
configuration is platform-dependent: https://help.github.com/articles/dealing-with-line-endings/#global-settings-for-line-endings

lhein:
Maybe we should create a .gitattributes files per repository which handles the correct line ending?

apupier:
we will need to investigate how to configure the .gitattributes for the repository.
This PR seems to fix my issue after i recreated the other PR: https://github.com/apupier/fuseide/pull/4

PR #765, opened 2016-11-18
FUSETOOLS-2160 - Allow move nodes between containers

apupier:
current limitation:
- doesn't support the insert/append/prepend

I think that the current state is coherent and is adding nice value.
I will implement the insert/append/prepend in another PR


editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/features/create/ext/CreateFigureFeature.java
```diff
@@ -350,8 +359,9 @@ protected String getIconName() {
 			// simple drop on a container
 			node = createNode(selectedContainerElement, selectedContainerElement != null);
 		}
+		node.initialize();
```
lhein: Hard to see from this diff but can node be ever null?

lhein: I just wondered because there is no null check.

PR #772, opened 2016-11-28
FUSETOOLS-2023 - Remove namespace example in When xpath expressions

apupier:
due to Camel bug https://issues.apache.org/jira/browse/CAMEL-10509

avoid usage of namespace in When XPath to avoid users to hit the Camel
bug when using a provided template


bfitzpat:
Unfortunately I don't have enough knowledge of this part of Camel to help on this one. @lhein can you take a look?

PR #776, opened 2016-11-29
FUSETOOLS-2132 - Use latest xsd available in cache

apupier:
The goal is to reduce probability of failing test (and also the
experience of the user which would have to do a refresh) due to bad
retrieval of latest version on internet

- update to 2.17.3 for most of the xsd
- update to 2.17.2 for camel-spring-integration which is the latest
available
- provide the big osgi blueprint xsd as cache
- kept activemq xsds as Camel is still using 5.11 version


apupier:
I don't understand what you mean. These cached xsds are used by Eclipse WTP to validate xml files that are declaring xsds. As camel don't have the precise version in the namespace we can only provide the latest one and let user precise a more precise one if needed.

lhein:
Yep, you are correct. My fault.

apupier:
no problem, always better to ask more questions than needed instead of less than needed :-)

PR #778, opened 2016-11-30
FUSETOOLS-2064 Change the order of Runtimes in Fuse Project wizard

vcornejo:
I was changed the LinkedHashMap to TreeMap using Natural Ordering instead of the original ordering.

### DIFF
--- a/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/FuseIntegrationProjectWizardRuntimeAndCamelPage.java
+++ b/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/FuseIntegrationProjectWizardRuntimeAndCamelPage.java
@@ -310,7 +310,7 @@ public class FuseIntegrationProjectWizardRuntimeAndCamelPage extends WizardPage 
 			runtimesSet = RuntimeManager.getRuntimes(Collections.singleton(facetVersion));
 		}
 		
-		Map<String, IRuntime> runtimesMap = new LinkedHashMap<String,IRuntime>();
+		Map<String, IRuntime> runtimesMap = new LinkedHashMap<>();
 		for (org.eclipse.wst.common.project.facet.core.runtime.IRuntime r : runtimesSet) {
 			IRuntime serverRuntime = FacetUtil.getRuntime(r);
 			if (serverRuntime != null) {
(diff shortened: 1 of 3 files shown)


editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/FuseIntegrationProjectWizardRuntimeAndCamelPage.java
```diff
@@ -261,27 +275,28 @@ public void run() {
...
-		runtimeCombo.add(Messages.newProjectWizardRuntimePageNoRuntimeSelectedLabel);
-		runtimeCombo.select(0);
+		runtimeComboViewer.setInput(null);
+		runtimeComboViewer.add(Messages.newProjectWizardRuntimePageNoRuntimeSelectedLabel);
+		final ISelection selection = new StructuredSelection(Messages.newProjectWizardRuntimePageNoRuntimeSelectedLabel);
+		runtimeComboViewer.setSelection(selection);
 		for (Map.Entry<String, IRuntime> entry : serverRuntimes.entrySet()) {
-			runtimeCombo.add(entry.getKey());
-			++i;
+			runtimeComboViewer.add(entry.getKey());
```
apupier: Optional: When using Viewers, we should not add directly entries to it.
There is setInput and then to calculate what is available, it is the role of the ContentProvider.
I won't block the PR for this point as it is still better than previously but if you can have a look and try to implement it following the state of the art it would be better.

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/FuseIntegrationProjectWizardRuntimeAndCamelPage.java
```diff
@@ -293,7 +310,7 @@ private void configureRuntimeCombo() throws CoreException {
 			runtimesSet = RuntimeManager.getRuntimes(Collections.singleton(facetVersion));
 		}
 		
-		Map<String, IRuntime> runtimesMap = new LinkedHashMap<>();
+		Map<String, IRuntime> runtimesMap = new LinkedHashMap<String,IRuntime>();
```
apupier: optional:  why did you add the generic types? It is simpler to stick with diamond operator new LinkedHashMap<>()

apupier:
testPR

apupier:
testPR

PR #784, opened 2016-12-05
FUSETOOLS-2213 - Upgrade Target Platform to 4.4.1.CR1

apupier:
- upgrade TP
- also downgrade commons.io dependency to 2.2 as 2.4 as been removed


apupier:
testPR

PR #788, opened 2016-12-14
FUSETOOLS-2216 - Basic support of list of elements for an eip not at

apupier:
first level

- current limitation is that elements can not contains a "," character
(still better than corrupting the data everytime before)
- in eips provided, no use case detected in which a "," can be used.
- UIs is providing a simple text comma separated

### DIFF
--- a/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
+++ b/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
@@ -13,9 +13,11 @@ package org.fusesource.ide.camel.model.service.core.model;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
@@ -40,6 +42,8 @@ import org.w3c.dom.NodeList;
  */
 public abstract class AbstractCamelModelElement {
 
+	private static final String NODE_KIND_VALUE = "value";
+	private static final String DESCRIPTION_NODE_NAME = "description";
 	public static final String NODE_KIND_EXPRESSION = "expression";
 	public static final String NODE_KIND_ATTRIBUTE = "attribute";
 	public static final String NODE_KIND_ELEMENT = "element";
@@ -173,16 +177,15 @@ public abstract class AbstractCamelModelElement {
 	 * @return
 	 */
 	public AbstractCamelModelElement findEndpoint(String uri) {
-		if (getChildElements().isEmpty()) {
-			if (getParameter(URI_PARAMETER_KEY) != null && ((String)getParameter(URI_PARAMETER_KEY)).equals(uri)) {
-				return this;
-			}
-		} else {
-			for (AbstractCamelModelElement cme : getChildElements()) {
-				AbstractCamelModelElement e = cme.findEndpoint(uri);
-				if (e != null) {
-					return e;
+		if (uri != null) {
+			if (getChildElements().isEmpty()) {
+				if (uri.equals(getParameter(URI_PARAMETER_KEY))) {
+					return this;
 				}
+			} else {
+				return getChildElements().stream()
+						.filter(cme -> cme.findEndpoint(uri) != null)
+						.findFirst().orElse(null);
 			}
 		}
 		return null;
@@ -376,7 +379,7 @@ public abstract class AbstractCamelModelElement {
 	 */
 	public void setDescription(String description) {
 		this.description = description;
-		setParameter("description", description);
+		setParameter(DESCRIPTION_NODE_NAME, description);
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/AbstractCamelModelElement.java
```diff
@@ -760,32 +763,54 @@ protected void setParameter(String name, Object value, boolean overrideChangeChe
 						break;
 					}
 				}
-			} else if (kind.equalsIgnoreCase("value")) {
+			} else if (kind.equalsIgnoreCase(NODE_KIND_VALUE)) {
 				e.setTextContent(null);
```
lhein: maybe reverse kind and constant

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/properties/FusePropertySection.java
```diff
@@ -674,6 +675,23 @@ public void modifyText(ModifyEvent e) {
...
+			//TODO: use something better than a text for a list of elements as we have the obvious limitation of not supporting the ","
+			List<String> initialValue = (List<String>)camelModelElement.getParameter(p.getName());
+			String delimiter = ",";
+			String collect = initialValue.stream().collect(Collectors.joining(delimiter));
+			Text txtField = getWidgetFactory().createText(parent, collect, SWT.SINGLE | SWT.LEFT);
+			txtField.addModifyListener(new ModifyListener() {
+				@Override
+				public void modifyText(ModifyEvent e) {
+					Text txt = (Text) e.getSource();
+					camelModelElement.setParameter(p.getName(), Arrays.asList(txt.getText().split(delimiter)));
```
lhein: Shouldn't we take care of possible whitespaces between delimiter and values?

lhein:
I am wondering if we should replace the text field in the properties with a list which can hold strings together with manipulator buttons like new/edit/delete. 

Or alternatively a read-only list showing all headers and a button for opening a manipulating dialog which allows to add, remove and edit items from the list.

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/properties/FusePropertySection.java
apupier: We don't know if whitepaces can start or end one of the value or not. So I would prefer to let user take care of it

apupier:
> I am wondering if we should replace the text field in the properties with a list which can hold strings together with manipulator buttons like new/edit/delete.

Would be clearly better.
I didn't want to implement it here for several reasons:
- not easy to provide that it in a usable way given that we are in the properties view and there are a lot of other fields
- this is closer to a new feature development than a bugfix
- wanted to ensure to provide a version which is unblocking our users

What do you think to merge this PR which is unblocking the client and open another JIRA issue with lower priority to handle this use case?

> Or alternatively a read-only list showing all headers and a button for opening a manipulating dialog which allows to add, remove and edit items from the list. In case of the CSV headers it might be even worth to think about a autodetect option where u can specify a CSV file and then let the user pick the delimiter and autodetect the header values.

This is clearly a complete new feature. For sure this one requires another JIRA and i think that the priority will not put it in 9.1.0.

apupier:
@jpav Can you review this PR? What is you opinion on Lars and mine comments?

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/properties/FusePropertySection.java
jpav: I can see both sides of this, but unless there's any example someone could come up with involving the need to have leading or trailing spaces around a value in a list of property values, then it seems like trimming those to avoid bugs would outweigh the flexibility of leaving it to the user/developer to filter them out when unintentional. I.e, it seems like their unintentional entry could be fairly common vs. their need in leading/trailing a value, which I suspect is extremely rare, if not non-existent.

core/tests/org.fusesource.ide.camel.model.service.core.tests.integration/src/main/resources/withMarshalCsvHeaders.xml
```diff
@@ -0,0 +1,27 @@
...
+                    <header>user_id</header>
+                    <header>ex_employee</header>
+                    <header>post_rank</header>
+                    <header>job_title</header>
+                </csv>
+            </marshal>
+            <to id="to_local_file" uri="{{local.generated.file.path}}?fileName=cx_export.csv.${date:now:yyyy-MM-dd-HHmmss}&amp;tempPrefix=."/>
+              </route>
+    </camelContext>
+</blueprint>
```
jpav: Maybe I'm missing something here, but where do the commas separating values show up in the test data?

apupier: in the file there is no comma, it is only a quick fix for the list of values inside the header fields

jpav:
I agree with both your comments about a different approach for entering these values, and their impact on this issue's status and timeframe, pushing it beyond 9.1.0.  Seems like another enhancement would be in order. This ties into the conversation we had in our last weekly meeting where we were discussing pop-up dialogs in lieu or in addition to what's offered in the properties view.  I think what @lhein suggests would more aptly go into such a dialog, and my personal opinion is that we should be moving away from use of the properties view altogether.

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/properties/FusePropertySection.java
apupier: if rare, it means that we introduce a loss of data to the user
if not existent, that's better to trim.
As we don't know i prefer to avoid the loss of data

lhein:
The only negative thing I saw is that the UI is not responsive enough when typing in the headers field. It seems to delay and work hard under the hoods with every character I enter in the list.

apupier:
I created https://issues.jboss.org/browse/FUSETOOLS-2233 for the follow-up

PR #794, opened 2016-12-22
JBTIS-1008 - Upgrade TP to include OpenShift Tooling

apupier:
- upgrade to 4.4.1.CR2-SNAPSHOT
- update Contributing.md: the Target Platform configuration in
Development environment need to be done in a different way, it requires
to use the "installation" instead of "Software site". Please note that
using "Directory" is not working because a plugin need to be unpacked
- updated the launch configuration for Windows


lhein:
If a plugin needs to be extracted to work then it should define this in its manifest.

apupier:
it is defined in the manifest but Eclipse is not respecting it when using "directory" for creating the Target Platform and when using the "Software site", several errors are reported.

lhein:
Ah... Did not know about that limitation.

Am 22.12.2016 3:49 nachm. schrieb "Aurélien Pupier" <
notifications@github.com>:

apupier:
oups, I thought that it was working but in fact not. I think that I had an already running cdk when i tried

PR #795, opened 2016-12-23
FUSETOOLS-2153 - Provide pre-configured Launch Configurations

apupier:
- one for launching locally
- one for deploying on OpenShift with mandatory java system properties
pre-filled.

those Launch Configuratiosn are available when doing Run As -> Run configuration...
image
for the local one:
image
For the deployment on OpenShift using fabric8-maven-plugin:
image
image

- VM arguments are used because fabric8-maven-plugin is not supporting maven properties and environment variables seems to not working neither


apupier:
I updated the name of the local launch configuration to "Launch locally <projectName>"

PR #799, opened 2017-01-04
FUSETOOLS-2228 - ignore parameters in Bundle-SymbolicName

lhein:


servers/tests/org.fusesource.ide.server.karaf.core.tests/src/test/java/org/fusesource/ide/server/karaf/core/server/subsystems/KarafUtilsTest.java
```diff
@@ -37,6 +39,13 @@
 	public TemporaryFolder testFolder = new TemporaryFolder();
 	
 	@Test
+	public void testGetBundleSymbolicNameWithoutParameters() throws CoreException {
```
apupier: isn't this test testing "with" and not without?
method names says without

servers/tests/org.fusesource.ide.server.karaf.core.tests/src/test/java/org/fusesource/ide/server/karaf/core/server/subsystems/KarafUtilsTest.java
```diff
@@ -37,6 +39,13 @@
...
+		assertThat(KarafUtils.stripParametersFromSymbolicName("mySampleApp;blueprint.aries.xml-validation:=false").indexOf(';')).isEqualTo(-1);
+		BundleDetails bd = new BundleDetails("mySampleApp;blueprint.aries.xml-validation:=false", "1.0.0");
```
apupier: better to write also a different test checking when there is no parameter

servers/tests/org.fusesource.ide.server.karaf.core.tests/src/test/java/org/fusesource/ide/server/karaf/core/server/subsystems/KarafUtilsTest.java
```diff
@@ -37,6 +39,13 @@
...
+		assertThat(bd.getSymbolicName().indexOf(';')).isEqualTo(-1);
```
apupier: why not checking assertThat(bd.getSymbolicName()).isEqualTo("mySampleApp"); ?
would be more precise test and easier to read, what do you think?

servers/tests/org.fusesource.ide.server.karaf.core.tests/src/test/java/org/fusesource/ide/server/karaf/core/server/subsystems/KarafUtilsTest.java
lhein: Get the symbolic name without parameters...thats what it tests

servers/tests/org.fusesource.ide.server.karaf.core.tests/src/test/java/org/fusesource/ide/server/karaf/core/server/subsystems/KarafUtilsTest.java
lhein: agreed

servers/tests/org.fusesource.ide.server.karaf.core.tests/src/test/java/org/fusesource/ide/server/karaf/core/server/subsystems/KarafUtilsTest.java
lhein: ok

servers/tests/org.fusesource.ide.server.karaf.core.tests/src/test/java/org/fusesource/ide/server/karaf/core/server/subsystems/KarafUtilsTest.java
apupier: ok, it depends to what we report the "withoutParameters" I had in mind that it was the case we are testing and not the method name
with the other tests, it will be clearer :-)

PR #806, opened 2017-01-11
FUSETOOLS-2227 - Add support for FUSE 6.3.0.R1

apupier:
- corresponds to Camel 2.17.0.630224


apupier:
timeout seems to have occured due to a timeout accessing artifacts from a staging repoisotry

PR #808, opened 2017-01-11
FUSETOOLS-2239 FUSETOOLS-2237 mid-air collision

apupier:
the way to calculate the AMQ version need to be updated as the latest
version is now a productized one


apupier:
it is testing that the dependencies version added for ActiveMQ component is correct.

Previously in the test, it was hardcoded to expect a community version as the "last camel version". WWith last PR on Master, it has moved to a productized version, so we need to test against a productized version.

PR #812, opened 2017-01-16
FUSETOOLS-2248 - Fix Camel version detection

apupier:
it was using the toString on the Selection instead of using the content
of the selection itself

providing the PR without test to check that i broke nothing else (and if someone want to write test)
I think that the "easiest" is to mock runtimes rturned by getServerRuntimes


apupier:
the test will be written by QE using UI tests, see https://issues.jboss.org/browse/JBTIS-1017

apodhrad:
Verified manually and it works

PR #813, opened 2017-01-17
FUSETOOLS-2252 - avoid touching pom at wrong time

apupier:
re-enable test for FUSE 6.3.0.R1 too(FUSETOOLS-2251)

want to see results on a clean environment


apupier:
ready for review

PR #816, opened 2017-01-18
Upgrade to 4.4.2.Final-SNAPSHOT

apupier:
upgrading to corresponding JBTIS TP used, using SNAPSHOT because the non-snapshot artefact is not available on repositories


apupier:
testPR

PR #818, opened 2017-01-18
FUSETOOLS-2253 - fix NPE when selecting method in transform

lhein:


editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/properties/FusePropertySection.java
```diff
@@ -432,7 +432,9 @@ protected void languageChanged(String language, Composite eform, AbstractCamelMo
 				this.selectedEP.setParameter(prop.getName(), expContainerElement);
 				uiExpressionElement = expressionElement;
 
-				updateDependencies(lang.getDependencies());
+				if (lang != null) { // some languages are not defined in catalog like "method"
```
apupier: Do you mean that the catalog is incomplete?
Shouldn't we report it to Camel, backport the fix to 2.18.1 FIS branch and regenerate the catalog?
or at least report it to Camel and fix the catalog manually?

lhein: I think method is not a real language. That's the whole point. But yeah...probably Camel should do something about that in terms of Catalog completeness. @davsclaus

lhein:
> If "method" is not a language how can it be listed in the combo?
Because its a defined One-Of value I guess.

lhein:
testPR

lhein:
testPR

apupier:
when using the "method" language, attributes are now correctly updated and initialized except the checkbox "Trim" which is not correctly initialized.

apupier:
seems that some "languages" are defined in eips... : https://github.com/fusesource/fuseide/blob/master/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/properties/FusePropertySection.java#L470
we cna effectively find the "method" in eips catalog:  https://github.com/fusesource/fuseide/blob/master/core/plugins/org.fusesource.ide.camel.model.service.v2_18_1_redhat_000007/catalogs/springboot/eips.xml#L2331

lhein:
Yes, it already looks in EIPs catalog if its not in Language catalog. So nothing to do about this really.

apupier:
- I tested manually and it is working fine now.
- I think we shoudl raise an feature request to QE in order that they write a UI tests, witthe curren tstate of -our codebase on this part I think that it will be hard to automate it differently.
- would worth digging why we don't have the "method" not declared in "languages", if it is in our internal catalog model or in Camel one that there is an issue

lhein:
testPR

PR #821, opened 2017-01-19
Avoid NPE in tests

apupier:
- report Marker without severity as Error and as Warning to fail test
when it occurs


apupier:
testPR

apupier:
I think this improvement can wait 9.2

lhein:
I agree

nickboldt:
testPR

apupier:
testPR

PR #826, opened 2017-01-24
FUSETOOLS-2262 - fix camel configurator

bfitzpat:
-- use older dependency method to get dependencies in another way from
the project

### DIFF
--- a/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/maven/CamelProjectConfigurator.java
+++ b/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/maven/CamelProjectConfigurator.java
@@ -280,6 +280,14 @@ public class CamelProjectConfigurator extends AbstractProjectConfigurator {
                 }
             }
         }
+        // use deprecated dependency method as a last resort 
+        if (mavenProject.getCompileDependencies() != null) {
+        	for (Dependency dep : mavenProject.getCompileDependencies()) {
+        		if (isCamelDependency(dep)) {
+        			return dep.getVersion();
+        		}
+        	}
+        }
         return CamelModelFactory.getLatestCamelVersion();		
 	}
 


bfitzpat:
I do not like this approach, but it's the only method I've found that actually successfully returns the dependency. @apupier or @lhein I am open to suggestions.

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/maven/CamelProjectConfigurator.java
```diff
@@ -280,6 +280,16 @@ private String getCamelVersion(MavenProject mavenProject) throws CoreException {
                 }
             }
         }
+        // use deprecated dependency method as a last resort 
+        if (mavenProject.getCompileDependencies() != null) {
+        	Iterator<Dependency> depIter = mavenProject.getCompileDependencies().iterator();
```
apupier: to stay consistent with the code in the same method, what do you think to avoid usage of an explicit Iterator and use for loop on the list from mavenproject.getCompileDependencies() ?
(it is reducing the number of  lines also)

bfitzpat: I'll look at that. Good suggestion. Thanks.

bfitzpat:
This needs to make it into this release due to the fact that we still must support older versions of SwitchYard in the latest version of the tooling. Not sure if that means a new CR or what.

apupier:
it means:
- a new CR
- updating the severity to "blocker" or "critical"
- be approved by QE as "blocker" or "critical" by QE

bfitzpat:
@apodhrad What do you think? Is this a blocker? Or can this wait to the next "dot" release of the tooling?

apodhrad:
The original issue is set as critical - https://issues.jboss.org/browse/SWITCHYARD-2989

apodhrad:
I will verify the PR and once I'm done I will elt you know whether to merge the PR

bfitzpat:
thanks @apodhrad

apodhrad:
Now the Camel facet is set correctly, but the error
```
Camel 2.18.1.fuse-000025 requires Java 1.8 or newer.
```
pups up again if you do a change in the camel editor. For example
1. Create a SY project with Java 7
2. Open switchyard.xml and add a Camel XML component
3. Open the Camel component (the Camel editor should be opened)
4. Add new component Set Body
5. Set the language, e.g. to constant
6. The error pops up

bfitzpat:
@apupier or @lhein  any ideas on this new wrinkle?

apupier:
I created https://github.com/fusesource/fuseide/pull/830 to continue the work started by Brian

apupier:
it has been handled on https://github.com/fusesource/fuseide/pull/830

PR #827, opened 2017-01-27
FUSETOOLS-2260 - restore old naming of plugins / features

lhein:


lhein:
Yes and that's exactly what the PR does.

PR #830, opened 2017-01-31
Fusetools 2262 handle camelversion not supported exactly

apupier:


apupier:
the remaining issue is that when adding a Language or DataFormat, the dependency will be the one from the closest compatible camel version and not the exact one specified in the pom.xml for camel-core

apodhrad:
I tried it locally and it seems to be working

apupier:
testPR

apupier:
testPR

PR #833, opened 2017-02-01
Initial check-in of new transformation runtime engine

jpav:


apupier:
after 1 minute looking at the PR here is my feedback:
- there is no JIRA associated
- spaces instead of tabs are used
- there are commented out sysout
- there are several empty javadoc comments
- date in header is 2015
- I don't get why a whole log mechanism seems to be rewritten
- does this code require change in Runtime too?
- where come the name chrysalix?
- package name is org.jboss, we have still org.fusesource (unfortunately) but if we change it, maybe we should use org.redhat

PR #834, opened 2017-02-02
Revert "Initial check-in of new transformation runtime engine"

apupier:
Reverts fusesource/fuseide#833

 I revert this PR, we are in CR phase only blocker and critical can be merged. Furthermore, there is a lot of code, it will take several days to review it.


lhein:
+1

No unreviewed code should make it into our code base and in the current phase it is to risky at all.

PR #836, opened 2017-02-02
FUSETOOLS-2222 - Use Camel Productized version for FIS 2.0 Final

apupier:


apupier:
due to timeout when downloading for the first time springboot dependencices, the build will surely need to be launched 2 or 3 times before it works

PR #838, opened 2017-02-06
FUSETOOLS-2276 - Stabilize test

apupier:
avoid usage of Thread.sleep and rely on Job instead


apupier:
The job was failing sometimes on the CI (typically if the CI was very slow). Using a more precise condition will avoid this issue. So faster when C is faster and test not failing when CI slower.
ok to rename

PR #857, opened 2017-02-21
FUSETOOLS-2268 - Upgrade parent pom to 4.4.3.Final

apupier:
it introduces a change in qualifiers, it is now using the latest git
commit date instead of the time the build occured

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -14,7 +14,7 @@
   <parent>
     <groupId>org.jboss.tools</groupId>
     <artifactId>parent</artifactId>
-    <version>4.4.0.Alpha1-SNAPSHOT</version>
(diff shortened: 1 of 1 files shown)


lhein:
yeah, because we did not upgrade yet to latest JBTIS TP 
jbtis.version = 4.4.2.AM1-SNAPSHOT

apupier:
I think that we don't need to follow that closely the jbtis version and using non-snapshot version is enough

lhein:
we have to....thats the version for the next release

apupier:
updated with
jbtis.version = 4.4.2.AM1 4.4.3.Final for parent pom

PR #874, opened 2017-03-02
FUSETOOLS-2304 - More precise search of Camel Editor

apupier:
- it should avoid NPE as the editor is retrieved from FeatureProvider
currently in use instead of the "potentially active one"

- factorize code to retrieve PictogramElement from Context of the
request
- replace spaces by tabs


apupier:
note: to review PR more easily you can ignore whitespace change by adding w=1 parameter, so it would be https://github.com/fusesource/fuseide/pull/874/files?w=1

PR #878, opened 2017-03-08
added jacoco profile

lhein:


pom.xml
```diff
@@ -1,433 +1,472 @@
...
+				<groupId>org.eclipse.tycho</groupId>
+				<artifactId>tycho-surefire-plugin</artifactId>
+				<version>${tychoVersion}</version>
+				<configuration>
+					<useUIThread>true</useUIThread>
+					<useUIHarness>true</useUIHarness>
+					<includes>
+						<include>**/*IT.class</include>
+					</includes>
+					<argLine>${tycho.testArgLine} -XX:+HeapDumpOnOutOfMemoryError
```
apupier: this should stay on the same line. I think that on recent version of Tycho, w can provide several tags to allow writing on different lines

pom.xml
```diff
@@ -1,433 +1,472 @@
...
+			<id>osx</id>
+			<activation>
+				<property>
+					<name>java.vendor.url</name>
+					<value>http://www.apple.com/</value>
+				</property>
+			</activation>
+			<properties>
+				<!-- THE FOLLOWING LINE MUST NOT BE BROKEN BY AUTOFORMATTING -->
+				<systemProperties combine.children="append"> -Dosgi.arch=x86
```
apupier: this should stay on the same line. I think that on recent version of Tycho, w can provide several tags to allow writing on different lines

lhein:
testPR

PR #881, opened 2017-03-13
Fix SonarQube test code coverage

apupier:


apupier:
tested locally, seems better (even without all tests) :-)
image

PR #882, opened 2017-03-13
WIP FUSETOOLS-2155

apupier:
- Provide Remote Camel Debug (without java because the java can be not enabled and only JMX is available)
- Provide Remote Camel and Java debug
- Clean code
- Provide disconnection only for remote process to be coherent with Java
Debugger

--> missing tests
but an initial feedback is welcome


apupier:
I splitted the work amongst several JIRA tasks

PR #890, opened 2017-03-16
FUSETOOLS-2326 - Avoid NPE when there are Linked resource in project

apupier:


editor/plugins/org.fusesource.ide.project/src/org/fusesource/ide/project/providers/CamelVirtualFolder.java
```diff
@@ -94,52 +94,48 @@ public void resourceChanged(IResourceChangeEvent event) {
...
-				try {
-					if (resource instanceof IFile && new CamelFilesFinder().isFuseCamelContentType((IFile) resource)) {
-						camelFiles.add(resource);
+			int deltaKind = delta.getKind();
+			if(deltaKind == IResourceDelta.ADDED) {
+				if (!camelFilesFinder.isWorkProjectFolder(project, getAncestorDirectChildOfProject(project, resource))) {
+					// a resource was added, check if we need to add it the the
+					// camel virtual folder too
+					try {
+						if (resource instanceof IFile && new CamelFilesFinder().isFuseCamelContentType((IFile) resource)) {
```
apupier: it is already less than it was before :(

editor/tests/org.fusesource.ide.project.tests.integration/src/main/java/org/fusesource/ide/project/providers/CamelVirtualFolderIT.java
```diff
@@ -34,6 +39,17 @@ public void testCamelVirtualFolderFindCamelFile() throws Exception {
 	}
 	
 	@Test
+	public void testCamelVirtualFolderDetectRemovedCamelFile() throws Exception {
```
apupier: for a test, I think that it is better to throw all Exceptions

PR #899, opened 2017-03-22
Remove useless assignments to local variable

apupier:


transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/extensions/TransformationDblClickHandler.java
```diff
@@ -73,19 +73,17 @@ public boolean canHandle(AbstractCamelModelElement clickedNode) {
 	 * @return	the parameter value or null
 	 */
 	private String getEndpointParameter(String uri, String key) {
```
apupier: this is used:
image

apupier:
MAJOR NewCamelTestWizardPageTwo.java#L400: Remove this unused private "collectEndpoints" method.  is used
image

PR #900, opened 2017-03-22
FUSETOOLS-2236 - Add Hystrix Routing EIP in the Palette

apupier:
"On Fallback" has been added for Hystrix


apupier:
wrong bug number

PR #902, opened 2017-03-22
Remove conditions which always evaluate to same value

apupier:


apupier:
MAJOR FusePropertySection.java#L553: Remove this unused private "updateDependencies" method.

false positive

image

PR #905, opened 2017-03-24
Fix primitives boxed just for String

apupier:
it improves performance


apupier:
false positive reported by Sonar:

image

PR #908, opened 2017-03-24
Avoid usage of old synchronized classes

apupier:
which should improve performance


jpav:
StringBuffer is synchronized, StringBuilder is not.

PR #913, opened 2017-03-28
use StringBuilder to concatenate String in Loop

apupier:


jpav:
Yes, because it avoids having to create new StringBuilders/StringBuffers and Strings under the covers (Not sure anymore whether the Java compiler has been updated use StringBuilders internally (which are not synchronized, unlike StringBuffers))

PR #917, opened 2017-03-29
FUSETOOLS-2347: Fix issues in TransformTestWizardPage found by SonarQube

jpav:


transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/wizards/TransformTestWizardPage.java
```diff
@@ -484,12 +474,13 @@ private void doStatusUpdate() {
...
-                || fPackageStatus.getMessage().trim().isEmpty())) {
-            // override this particular case, since the default package is ok, though not great
-            fPackageStatus = new StatusInfo(NONE, null);
-        } else if (fPackageStatus.getCode() == StatusInfo.WARNING && fPackageStatus.getMessage() != null
-                && fPackageStatus.getMessage().contains("default package is discouraged")) { //$NON-NLS-1$
+        boolean defaultPkg = fPackageStatus.getCode() == StatusInfo.ERROR && (fPackageStatus.getMessage() == null
+            				 || fPackageStatus.getMessage().trim().isEmpty());
+        if (!defaultPkg) {
+        	defaultPkg = fPackageStatus.getCode() == StatusInfo.WARNING && fPackageStatus.getMessage() != null
+                		 && fPackageStatus.getMessage().contains("default package is discouraged"); //$NON-NLS-1$
```
bfitzpat: Can we externalize this string?

apupier: This String is not ours, but woudl be better to use the Externalized version from JDT

jpav: So, not sure what you want me to do here guys.  As @apupier pointed out, this isn't our string, and I also have no idea where to find a publicly accessible externalized version of it in the JDT.

bfitzpat: https://eclipse.googlesource.com/jdt/eclipse.jdt.ui/+/968ae67958a8fe138a4dd443b7b3f078bca5cd3e/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/wizards/NewWizardMessages.java#107

bfitzpat: Should be able to use org.eclipse.jdt.internal.ui.wizards.NewWizardMessages.NewTypeWizardPage_warning_DefaultPackageDiscouraged

That said, I'm not sure if it's better to just let it go than refer to an internal class in the JDT.

@apupier WDYT?

apupier: I think it is best to use the internal JDT class, otherwise it is not working when a user is using the Fuse Tooling with a Locale translated different than English.

jpav: I absolutely don't want to refer to an inner class

jpav: The whole point of externalization is to handle locale issues, I know, but it's also always a bad idea to use an inner class.  I guess I'll concur with whatever the majority vote is.  @lhein @bfitzpat?

apupier: better to have a working software using internal classes than a non-working software using none.
You can also change the logic to not rely on an  Externalized string and find some public api for that.

bfitzpat: Although I hate the idea of referencing internal classes, this is definitely not the only place that it happens and I do worry about localization (though I don't believe we've had any issues reported yet with localization and this is far from the only place where it comes up), I think referencing the inner class is better than having a brittle string that could change without warning down the line. So that's a wishy washy way of giving a +1 for the internal class though I hate the idea.

jpav: Okay, so looking into this more, changing the message regarding the default package is easy, but there's another case of this same issue just a few lines above with the container (i.e., folder) not existing.  As an aside, I looked for another API to accomplish this, and there isn't a direct non-private one, although I could effectively jump through some code hoops to reproduce the same logic.  But back to this other issue, the message takes an argument, so I have no easy way to compare the formatted output without recreating that argument.  So, the only option I see is to basically override the containerUpdated() method, which is not small, and change its behavior to skip the cases we want ignored.  Does everyone think this is all worth it?  Keep in mind all I was trying to address were a couple of "simple" Sonar issues...

bfitzpat: Ugh and double ugh. I'm good with simply //$NON-NLS-1$ 'ing the message and calling it good for now. Without a simple, consistent way of handling it I'm not sure what else to do in the short term.

jpav: Something else to keep in mind: this entire effort to reuse the Java New Type wizard to create a test class will be obsolete anyways once we migrate to using the Media Driver engine and data preview. So, @apupier?, and objections to approving this now?

PR #918, opened 2017-04-03
FUSETOOLS-2318 - Make CamelDebugFacade Jolokia compliant

apupier:
- add test to ensure it
- cast to Collections instead of HashSet

Need to wait to have a Target Platform which is including a newer version of Jboss Tools Server (especially this PR https://github.com/jbosstools/jbosstools-server/pull/490 ) before the test can work and so be able to merge


lhein:
PR has merge conflict

apupier:
testPR

PR #923, opened 2017-04-07
FUSETOOLS-2356 - Avoid stack in log for non local IFile

apupier:
bad side is that it is using deprecated API but there is no replacement
for it and this is used in getContentDescription so no other choice
(except ignoring the exception but it is even uglier from my point of
view)

Other opinions?


core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/util/CamelFilesFinder.java
```diff
@@ -98,7 +98,8 @@ private boolean isTestProjectFolder(IProject project, IResource f) {
 	 */
 	public boolean isFuseCamelContentType(IFile ifile) throws CoreException {
 		if( ifile != null
-				&& ifile.isSynchronized(IResource.DEPTH_ZERO)){
+				&& ifile.isSynchronized(IResource.DEPTH_ZERO)
+				&& ifile.isLocal(IResource.DEPTH_ZERO)){
```
apupier: it is effectively deprecated but it is the only to avoid the stack in log as the deprecated method is still in use in Eclipse internals

PR #924, opened 2017-04-10
FUSETOOLS-2315 - Use relative uri to ensure working in built version

apupier:
I referenced the xsd files in extension points as it is done in other places. in JBoss Tools and it is fixing the issue.

Previously it was working only in environment mode.

I don't explain why it was ok for a part of the xsd validation and not for some atttributes.

### DIFF
--- a/editor/plugins/org.fusesource.ide.catalogs/plugin.xml
+++ b/editor/plugins/org.fusesource.ide.catalogs/plugin.xml
@@ -5,31 +5,31 @@
     <catalogContribution id="default">
       <uri
           name="http://activemq.apache.org/schema/spring"
-          uri="platform:/plugin/org.fusesource.ide.catalogs/xsd/fuse/activemq-spring-5.11.0.redhat-620133.xsd" />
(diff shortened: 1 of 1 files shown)


apodhrad:
I'm still getting the error as described at FUSETOOLS-2315

apupier:
@apodhrad can you provide more details on how you tested it?
are you sure that you had a single version of the plugin available? was the plugin correctly resolved?

apodhrad:
@apupier you are right - with the clean installation it works.

apupier:
@apodhrad so the fix is working? Just need to wait the PR review from a Fuse Tooling developer?

apodhrad:
Please wait for a while. I have encountered that sometimes it is working and sometimes not :( I will try to find out what's the reason.

apodhrad:
@apupier I think that your changes do not have any effect on the XML validation. I found that the problem is caused by the SwitchYard plugin.

sy-camel-spring-xsd

apupier:
https://github.com/jboss-switchyard/tools/pull/660 seems to be the correct solution

PR #938, opened 2017-04-14
FUSETOOLS-2378 - Rename archetype catalog

apupier:
the provided archetype catalog has been used up until Fuse 6.2.1
version, it is now on normal redhat GA repository


editor/plugins/org.fusesource.ide.branding/OSGI-INF/l10n/bundle.properties
```diff
@@ -26,6 +26,6 @@ project.wizard.project.description = Create a new Fuse Project using the Maven b
 
 rider.preferences.label=Fuse Tooling
 
-remote.description = Fuse Tooling Remote Archetype Catalog
+remote.description = Fuse 6.2.1 and before Remote Archetype Catalog
```
lhein: Maybe rename to "Fuse 6.2.1 and earlier..."

PR #941, opened 2017-04-24
adapt version to 10.0

lhein:


lhein:
will create a clean set of PRs

PR #942, opened 2017-04-25
adapted to m2e 0.8.1

lhein:


lhein:
testPR

apupier:
> Also, why not set [0.8.1,0.10) ?

we might put [0.8.1,1.0.0) as some non-compatible API break has been introduced but there is no reason to restrict to 0.10 as there should be only additional API and nothing should be broken.

We usually don't put upper bound, I have no strong opinion on that, we are on an optimistic way of working but less robust

nickboldt:
If you're happy with being installed against 0.8.1+, with no upper bound set, then +1.

pom.xml
```diff
@@ -91,6 +91,12 @@
 	<repositories>
 
 		<repository>
+			<id>tychoM2E</id>
+			<layout>p2</layout>
+			<url>http://download.jboss.org/jbosstools/updates/requirements/m2eclipse-tycho/0.8.1.201704211436/</url>
```
apodhrad: Why do you specify this repository in jbosstools-fuse/pom.xml? The repository should be defined in the appropriate JBTIS-TP, shouldn't it?

lhein: That is a test. I am looking for a temp. workaround. The m2e stuff will move into the JBT TP anyways soon and therefore out of the JBTIS TP. Once that is done I will remove the definition....or alternatively if the addition did not help in my test :D

apodhrad: Ok, it makes sense ;-)

apupier: Can you create a JIRA with blocker state for 9.2.0 to think to remove this before we merge this PR?

lhein:
WOOOHOOOOOO!

PR #953, opened 2017-05-02
override the default surefire timeout

apupier:
rebased version of https://github.com/jbosstools/jbosstools-fuse/pull/950 as the 4.4.3.CR1-SNAPSHOT seems to have vanished


apupier:
testPR

PR #957, opened 2017-05-03
FUSETOOLS-2390 - fixing issue with create feature

bfitzpat:
(Not critical for 9.2, but would be good to go for 9.3 or 10.0)


lhein:
testPR

PR #958, opened 2017-05-04
FUSETOOLS-2386 - should fix the NPE

bfitzpat:
Good for 9.3 or 10. Haven't been able to reproduce the issue locally, but I suspect the file doesn't exist and that is the cause of the NPE.


lhein:
testPR

PR #945, opened 2017-04-25
FUSETOOLS-2263 - decouple camel version and use catalogs directly

lhein:

PR #981, opened 2017-06-12
FUSETOOLS-2283: New Fuse Integration Project wizard - an ugly rectangle

jpav:


jpav:
Jenkins is faling with this, but I have no idea why; apparently something to do with JMX based upon the error messages shown in the console.  Same type of thing happens locally, and it seems pretty consistent.  AFAICT, the actual issue has been fixed properly.  I don't think the test failure is related, but I'm over my head with whatever is happening there.  Can one of you guys more familiar with this take a look?

bfitzpat:
Known issue. @lhein  and @apupier have been looking into it for a while and now Nick is helping too I think.

jpav:
So it’s okay to be reviewed and checked in then, I assume?

bfitzpat:
Nope. Reviewed-- yes. Merged -- probably not until we get the build/test failures figured out locally and on Jenkins.

lhein:
the 2 launch files don't really belong into this PR. Furthermore I'd like to add that the wizard page you reworked will change with the decouple PR  #945. Please make sure that your fix is ported to that PR as well.

apupier:
> Furthermore I'd like to add that the wizard page you reworked will change with the decouple PR #945. Please make sure that your fix is ported to that PR as well.

I think that the merge will need to be done in the PR which will be merged the latest. I see no reason to block a merge for synchronization of branch.

lhein:
I did not block the PR because of the sync but because of the 2 launch files which don't belong there.

apupier:
testPR

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/FuseIntegrationProjectWizardRuntimeAndCamelPage.java
```diff
@@ -75,18 +67,19 @@
...
 		setDescription(Messages.newProjectWizardRuntimePageDescription);
 		setImageDescriptor(ProjectTemplatesActivator.imageDescriptorFromPlugin(ProjectTemplatesActivator.PLUGIN_ID, ProjectTemplatesActivator.IMAGE_CAMEL_PROJECT_ICON));
 		setPageComplete(false);
 	}
-	
+
 	/* (non-Javadoc)
 	 * @see org.eclipse.jface.dialogs.IDialogPage#createControl(org.eclipse.swt.widgets.Composite)
 	 */
+	@SuppressWarnings("unused")
```
apupier: this "SuppressWarnings" is reported as unused so it can be removed

jpav:
Apparently my Eclipse prefs were a bit different than yours.  I've changed that and removed the @SuppressWarnings.  Can someone approve this now?

apupier:
> Apparently my Eclipse prefs were a bit different than yours. I've changed that and removed the @SuppressWarnings.

it means that i have some Eclipse warnings not activated? Can you tell which warning do you have when you remove this @SuppressWarnings?

apupier:
@jpav are you on Mac? Would be nice to provide a screenshot of the new UI with your fix to easily compare with the old one (and to documen tit in the JIRA after)

jpav:
I am on a Mac.  I’ll include the screenshot in the JIRA.

jpav:
The warning is under Java > Compiler > Error/Warnings > Potential programming problems > Unused object allocation.  I had set to warning.

apupier:
> The warning is under Java > Compiler > Error/Warnings > Potential programming problems > Unused object allocation.  I had set to warning.

ok, i see the warning in this case but they seem not very relevant.
SonarLint is not reporting them neither. I think that it is better to ignore these ones. What do you think? Do you have some specific cases in which it is very useful?

jpav:
There has been a time or two when this has caught some defunct code, but I’m totally good with this being set to ignore.  I’ve already changed it to that on my side.

PR #982, opened 2017-06-12
FUSETOOLS-2332 - fixed issue where wasn't finding dataformat

bfitzpat:


lhein:
could you verify that this change is also ported to the decouple branch?

apupier:
> could you verify that this change is also ported to the decouple branch?

I think that the merge will need to be done in the PR which will be merged the latest. No reason to block a merge for synchronization of branch. A good way to ensure that the merge will be successful is to provide a test.

lhein:
Nobody blocked a branch Aurelien. I just said he should make sure the same patch is ported to the decouple branch. Its for our own benefit if we port those fixes for that branch otherwise it will become very messy for the person in charge for merging the decouple branch.

bfitzpat:
Wouldn't it be easier to cherry-pick the commit back to the decouple branch once it's committed to the master?

lhein:
I am not sure if that works. depends on how much that class has been changed and if that class still exists (DataFormatModel)

lhein:
In any case we need a Jira to track the port to decouple branch and link it to the decouple jira.

apupier:
Why not simply rebasing the "decouple branch" on master when it will be merged as we are doing for all other branches?

lhein:
Do u want to apply for that work? If yes then I am fine with it ;)

bfitzpat:
There's a handful of lines in this PR, so I wouldn't think the impact would be huge? I understand that the decouple branch is a beast, but I have regularly been updating the rebase the bean branch (also a beast) on the master branch to hopefully keep it up to date.

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/dataformats/DataFormatModel.java
```diff
@@ -72,7 +72,11 @@ public void setSupportedDataFormats(ArrayList<DataFormat> supportedDataFormats)
 	 * @return
 	 */
 	public DataFormat getDataFormatByName(String dataformat) {
-		return supportedDataFormats.stream().filter(df -> df.getName().equals(dataformat)).findFirst().orElse(null);
+		DataFormat foundDF = supportedDataFormats.stream().filter(df -> df.getName().equals(dataformat)).findFirst().orElse(null);
+		if (foundDF == null) {
+			foundDF = supportedDataFormats.stream().filter(df -> df.getModelName().equals(dataformat)).findFirst().orElse(null);
```
bfitzpat: I think getDataFormatByName is still accurate, but now it has a two-pronged approach. If name fails (which it seems to for data formats sometimes), we simply fall back to the model name. It only used getName() previously and this is limited to data formats, so my limited testing (and existing unit tests) have not raised any red flags yet.

I think getDataFormatByNameOrModelName, though more explicit, is an unnecessary step since the code is so simple at this point.

apupier: for dataformat which have several model names, you are taking arbitrarily the first one found. It means that we don't care which one is retrieved? How is it possible?

apupier: ok I understood the difference, for the name it is unique in the model by definition but the modelName can be shared amongst several dataformat

bfitzpat: I think we're relying on the "name" or "model name" being unique vs. the GAV. But I'm just basing that assumption on the functionality that was there previously.

apupier: name is unique but modelName is not unique

apupier: I feel that it is not clean to take the first dataformat which is found but it seems that it not causing any issue.

bfitzpat:
@apodhrad may have some ideas on how to test the dataformat code. It will affect any Camel component property sheet that uses Data Formats, so there's not a great way to test this all the way around.

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/dataformats/DataFormatModel.java
lhein: modelName is a single unique value per dataformat. Not sure what you are pointing at @apupier

apupier: I'm pointing that for an unique modelName there are several dataformat

lhein: Right. for instance for all 4 different types of the json dataformat (gson, jackson, xstream, johnzon) there is the modelName always json but the dataformatName is differing.

apodhrad:
I will definitely create a test for that. I already have such task in Jira so now it is on top of my todo list
https://issues.jboss.org/browse/JBTIS-1011

apodhrad:
Btw I've tried the PR and it seems to be ok but I'm still getting exceptions "Widget is disposed" whenever I expand the combo box for the first time (e.g. close, reopen the editor and expand the combo box)

bfitzpat:
@apodhrad Any chance you can share the exception and repro steps? I can take a look?

apodhrad:
@bfitzpat please have a look at https://issues.jboss.org/browse/FUSETOOLS-2425

bfitzpat:
Will hold this for a future sprint. Need to look into the test failures.

bfitzpat:
testPR

bfitzpat:
(Test that failed has nothing to do with the code that changed, so re-running build to see if whatever caused the failure shakes loose.)

bfitzpat:
Test failed - org.fusesource.ide.projecttemplates.tests.integration.wizards.FuseIntegrationProjectCreatorRunnableForOSESringBootIT.testOSESpringBootProjectCreation(FuseIntegrationProjectCreatorRunnableForOSESringBootIT.java:29)

@apupier Any ideas? Not a biggie right now but seems we've hit this before.

apupier:
know unstable test https://issues.jboss.org/browse/FUSETOOLS-2516

bfitzpat:
Got it. Thanks @apupier

lhein:
@bfitzpat to write the test missing

bfitzpat:
Rebasing the PR on master and will then work on the test...

bfitzpat:
Added new test if someone wants to take a look. I tested by hand in the tooling and I no longer see the NPE reported either. Also going to add a few more tests.

core/tests/org.fusesource.ide.camel.model.service.core.tests.integration/src/main/java/org/fusesource/ide/camel/model/service/core/tests/integration/core/model/DataFormatsInCamelModelElementIT.java
```diff
@@ -78,4 +79,12 @@ public void testDataFormatsInUnmarshal_notRemovedWhenUpdatingRefParameter() thro
 		assertThat(((AbstractCamelModelElement) unmarshallElement.getParameter("dataFormatType")).getNodeTypeId()).isEqualTo("json");
 	}
 
+	@Test
+	public void testCanFindTypes() throws CoreException, IOException {
```
bfitzpat: The problem here is that this tests finding data formats from the main CamelModel class, not the DataFormatModel. I have attempted to instantiate the DataFormatModel in a test, but it doesn't get populated correctly.

bfitzpat: Also, this test will fail for many of the model versions. Not quite sure how to filter out which ones will work and which ones won't.

bfitzpat:
Regardless of the test failing, this is not working correctly. It doesn't throw the NPE, but it's also not working in the drop-down. Will look further in the morning.

core/tests/org.fusesource.ide.camel.model.service.core.tests.integration/src/main/java/org/fusesource/ide/camel/model/service/core/tests/integration/core/model/DataFormatsInCamelModelElementIT.java
```diff
@@ -78,4 +79,12 @@ public void testDataFormatsInUnmarshal_notRemovedWhenUpdatingRefParameter() thro
...
+		String[] typesToFind = {"csv", "json", "string"};
```
lhein: I could be wrong but shouldn't it be more like 

`String[] typesToFind = {"bindy", "json"};`

And shouldn't we check for result set size > 1 ?

bfitzpat: http://joel-costigliola.github.io/assertj/core/api/org/assertj/core/api/AbstractIterableAssert.html#isNotEmpty()

bfitzpat:
Ok. Complete 180 as I went back to investigate this. We have changed paths to streamline the problem and are now using the tag name, which appears to be more unique. Now, the tests are not succeeding on all versions of Camel, so that needs to be addressed, but this works more consistently now in my local testing and seems to do the trick.

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/dataformats/DataFormatModel.java
```diff
@@ -1,103 +0,0 @@
-/*******************************************************************************
```
lhein: why did u delete that class?

bfitzpat: It's not used anywhere. Not in tests. Not in code. Nowhere.

lhein: wow, that seems to be a zombie from the xml model days....wondering how and why it survived.

bfitzpat: Not sure, but definitely not needed. :)

lhein:
This PR is not solving all problems.
Additional issues found:
- when selecting json / bindy dataformat the source doesn't get updated
- when selecting json / bindy dataformat the pom doesn't get updated correctly
- when switching the impl for json or bindy the pom doesn't get updated

I believe we are just scratching the surface with this PR. I would suggest to push this issue to the next release. We can't make it work good enough for this version. We should also rethink the dataformat combo box content. Maybe its easier to provide all entries for bindy / json there and then hiding the Library / Type parameter instead from the properties view.

bfitzpat:
Sounds like a plan.

PR #984, opened 2017-06-13
JBIDE-24557 disable running the...

nickboldt:
JBIDE-24557 disable running the tests.integration plugins from the main reactor so they can be run separately

### DIFF
--- a/all-itests/pom.xml
+++ b/all-itests/pom.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>org.fusesource.ide</groupId>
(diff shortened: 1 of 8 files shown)


nickboldt:
OK, that should be the last edit for this PR. I hope. :)

nickboldt:
What do you think of having a single pom to run all the tests in one step, without having to also build all the other plugins?

I can simplify this PR to just do that and reorg the site/category.xml file.

apupier:
> What do you think of having a single pom to run all the tests in one step, without having to also build all the other plugins?

i'm not sure to see the benefit.
when mvn without clean is launched, i think that it is not making a big difference in time if there is no change? (and if there are we are happy they are recompiled)
it means that there will be another pom to maintain? But if it is referencing all the parent tests pom maybe it is not a lot of more maintenance introduced.
if we want this functionality wouldn't it be best to have a profile for that? for instance putting all non-test in a profile which is active by default and that we can deactivate.

nickboldt:
Whether it's a collector pom or a profile, the information is the same... "here's a list of test plugins to build & run" .

If you prefer to house this information in your root pom, that's cool. I can create a "tests" profile which would list the test plugin modules to build & run.

The difference between an optional profile and a second collector pom is:
* if you enable the profile, you'll build everything AND the tests stuff too
* if you use a different pom, you can build & run ONLY the specific tests plugins (without re-building all the rest)

So it depends on if you want OR functionality, or AND functionality.

nickboldt:
close.

PR #983, opened 2017-06-13
JBIDE-24244 add tests to update site so they...

nickboldt:
JBIDE-24244 add tests to update site so they can be published to JBT coretests site; add all-tests/pom so we can run all the tests in a single step (without having to rebuild everything)

Signed-off-by: nickboldt <nboldt@redhat.com>

PR #989, opened 2017-06-21
JBIDE-24557 add tests to Fuse update site so...

nickboldt:
JBIDE-24557 add tests to Fuse update site so they can be aggregated by downstream JBT Tests site


nickboldt:
Note: this is a simpler PR than the previous one #983 which I've closed. If you like this the next step would be to refactor the test plugins so they generate source bundles, then to add those to the update site too.

PR #992, opened 2017-06-27
[DO NOT MERGE] Fusetools 2257 custom local repo for integration tests

apupier:


apupier:
This PR will help to investigate on [corrupted jars issue](https://issues.jboss.org/browse/FUSETOOLS-2392)
We should be able to detect how often we get corrupted jars from a fresh .m2
(but it is not fixing it, i think that we will need to have the mentioned issue fixed before applying this PR otherwise the old CI will be less "controlled")

apupier:
now a clean slave node is used everytime so we won't encounter the issue on CI anymore so closing the PR

PR #995, opened 2017-06-27
FUSETOOLS-2392 experiment with using better...

nickboldt:
FUSETOOLS-2392 experiment with using better nexus URLs to resolve dependencies; note that https://repo.fusesource.com/nexus/content/groups/ea redirects to repository.jboss.org so reuse that URL instead

### DIFF
--- a/editor/plugins/org.fusesource.ide.projecttemplates/template-sources/blank/blank-blueprint/pom.xml
+++ b/editor/plugins/org.fusesource.ide.projecttemplates/template-sources/blank/blank-blueprint/pom.xml
@@ -35,7 +35,7 @@
 		<repository>
 			<id>fuse-public-repository</id>
 			<name>FuseSource Community Release Repository</name>
-			<url>https://repo.fusesource.com/nexus/content/groups/public</url>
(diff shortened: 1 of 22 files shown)


nickboldt:
Built fine locally w/ -DskipTests=true. Now running again, with tests not skipped. Should be a couple hours before I have results. Meanwhile a PR build could happen in Jenkins :)

editor/plugins/org.fusesource.ide.projecttemplates/template-sources/blank/blank-blueprint/pom.xml
```diff
@@ -35,7 +35,7 @@
 		<repository>
 			<id>fuse-public-repository</id>
 			<name>FuseSource Community Release Repository</name>
-			<url>https://repo.fusesource.com/nexus/content/groups/public</url>
+			<url>${fusesourceNexusCommunity}</url>
```
apupier: these poms are not part of the build, they are templates. So unless you are replacing the value in the file during the build I don't think that it is working.

lhein:
I don't see the point in this PR. Aurelien pointed it out already. Closing the PR.

PR #1004, opened 2017-07-12
FUSETOOLS-2455 - fixing issue with global DF wizard hang

bfitzpat:


bfitzpat:
testPR

lhein:
for me it now freezes my UI every time I add a new dataformat which needs a download of libs

bfitzpat:
Yeah, apparently this doesn't actually fix the issue. Back to the drawing board. But I can't reproduce it on my end, so I'm open to suggestions. :)

apupier:
proposal to fix it in another PR: https://github.com/jbosstools/jbosstools-fuse/pull/1066

PR #1006, opened 2017-07-13
Fusetools 2456 fix technical debt reported by sonar o decouple camel version branch

apupier:


apupier:
I can create separated PR if you prefer

core/plugins/org.fusesource.ide.camel.model.service.impl/src/org/fusesource/ide/camel/model/service/internal/CamelService.java
```diff
@@ -54,8 +54,9 @@
 	public CamelService() {
 		try {
 			tempFolder = File.createTempFile("grape","m2repo");
-			tempFolder.delete();
-			tempFolder.mkdirs();
+			if(tempFolder.delete()){
```
lhein: So if the tempFolder does not yet exist, we do not create it any longer?

core/plugins/org.fusesource.ide.camel.model.service.impl/src/org/fusesource/ide/camel/model/service/internal/CamelService.java
```diff
@@ -261,8 +262,9 @@ public boolean isCamelVersionExisting(String camelVersion) {
 		MavenVersionManager tmpMan = new MavenVersionManager();
 		try {
 			File folder = File.createTempFile("grape", UUID.randomUUID().toString());
-			folder.delete();
-			folder.mkdirs();
+			if(folder.delete()) {
+				folder.mkdirs();
```
lhein: same as above. delete only returns true if the delete was successful. If the folder did not exist, the mkdirs will never be executed.

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/internal/Trace.java
```diff
@@ -1,48 +0,0 @@
...
-import org.eclipse.osgi.service.debug.DebugOptionsListener;
-import org.jboss.tools.foundation.core.plugin.AbstractTrace;
-
-/**
- * Helper class to route trace output.
- * The superclass keeps track of the debug options for your plugin, as well 
- * as when they change. The static methods in this class are here
- * only for convenience and easy of use
- */
-public class Trace extends AbstractTrace implements DebugOptionsListener {
```
lhein: Why did u remove the Trace? To have that seems to be defacto standard in JBT afair.

core/plugins/org.fusesource.ide.camel.model.service.impl/src/org/fusesource/ide/camel/model/service/internal/CamelService.java
apupier: effectively, that's not nice. thanks for catching it

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/internal/Trace.java
apupier: I checked in JMX plugins and they don't have it.
We are not using the Trace in our plugin so no need for it

core/plugins/org.fusesource.ide.camel.model.service.impl/src/org/fusesource/ide/camel/model/service/internal/CamelService.java
apupier: hum in fact, the file has been created the line before... which is really strange, why creating a file to delete and then create a folder?

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/internal/Trace.java
lhein: JMX is a bad source of information as it was copied over from fuseide ;) Better orientate on native plugins made by JBT team.

core/plugins/org.fusesource.ide.camel.model.service.impl/src/org/fusesource/ide/camel/model/service/internal/CamelService.java
lhein: right. that doesn't make much sense.

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/internal/Trace.java
apupier: Anyway, as we are not using it, I think it is better to remove it.

apupier: In Jboss tools doc, I understand that this is not mandatory to have it: https://github.com/jbosstools/jbosstools-devdoc/blob/a9c13b07fbfd1330cefca20ab0aef2f2dc56f84a/debugging/tracing_logging_in_eclipse.adoc

"Each plugin **can declare** its own trace keys specific to their needs"

apupier: on the other hand, https://github.com/jbosstools/jbosstools-devdoc/blob/ede3c46ae8c86592117e7bdcea14c53135d9dc72/api/foundation/foundation_api.adoc#conclusion it says that it is "highly recommended"

apupier: I still think that if not used, we should remove the code

bfitzpat: I agree we should get rid of that if we're not using it. It exists in the history, so we can get it back.

lhein: I still would prefer to keep trace as it gives us flexibility. Maybe split those file off into another pr to discuss later.

apupier:
put the "trace" dead code to another PR: https://github.com/jbosstools/jbosstools-fuse/pull/1018

PR #1007, opened 2017-07-13
FUSETOOLS-2420 - adapted the jacoco dest path variable for test purposes

lhein:


apupier:
the issue was coming from the usage of pack200

PR #1009, opened 2017-07-18
Fusetools 2422 clean code during investigation

apupier:


core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/CamelFile.java
```diff
@@ -495,4 +498,16 @@ public CamelModel getCamelModel(){
 
 		return result;
 	}
+
+	private List<AbstractCamelModelElement> findAllGlobalelementsWithId(String nodeId) {
```
apupier: There is no need for it now. I just extracted the method to have not too much nested if/for
And I think that for now, for the part that we are supporting, the id must be unique between camel elements and non-camel elements

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/CamelFile.java
```diff
@@ -153,14 +155,19 @@ public void initialize() {
 		}
 	}
 
+	private boolean isGlobalBean(Node child) {
```
apupier: yes you pointed it :-)
There is effectively a bigger issue.
We have a functionality which is allowing to simply list Global elements in the global editor even if they are not really known by the editor.
That's why the check is only for the GlobalBean element. it will avoid to have the Global bean property and wizard available if there is another "bean" tag name which is not a blueprint or spring bean.

The issue is that those other global elements can be:
- camel ones
- non-camel ones

but they are both handled the same way and the non-camel ones are forced to have an "id" attribute otherwise it will not work correctly.

I decided to not handle those external elements without id for now because no use case has been provided until now and it will require additional work. I created this enhancement: https://issues.jboss.org/browse/FUSETOOLS-2466

PR #1011, opened 2017-07-18
FUSETOOLS-2464 - clean code during investigation

apupier:


apupier:
I mainly followed recommendations from Sonar analysis so I take the action to merge it without further review.
Anyway, if there are some feedback on the modification, they can still be provided and I will have a look to them for sure.

PR #1012, opened 2017-07-18
FUSETOOLS-2464 - use String instead of Charset as Jaxb property

apupier:
Jaxb is supporting only String

need to be tested on a built environment (cannot go to the end with dev environment)


apupier:
Andrej has kindly tested the fix (see https://issues.jboss.org/browse/FUSETOOLS-2464?focusedCommentId=13437443&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13437443 )

bfitzpat:
Nice! Thanks @apodhrad !

PR #1013, opened 2017-07-19
JBIDE-24703 - Remove Integration Stack Target Platform bits

apupier:
- move up the jboss tools one
- update the jboss-tools target file for the latest versions available
- update readme on how to setup the environment


nickboldt:
after quick scan, looks reasonable. didn't try to build. not sure why you're moving stuff around but *shrug*.

apupier:
> not sure why you're moving stuff around

The new jbosstools-based Target Platform was added inside the JBTIS-based Target Platform. I moved up the jbosstools-based TP to avoid having nested folders.

PR #1016, opened 2017-07-20
FUSETOOLS-2460 - Fix data loss of Factory bean when model is reloaded

apupier:


bfitzpat:
I think we'll need to break the checkAllCommonAttributesLoaded test into two -- one for Spring and one for Blueprint -- but other than that I think this makes sense.

core/tests/org.fusesource.ide.camel.model.service.core.tests.integration/src/main/java/org/fusesource/ide/camel/model/service/core/tests/integration/core/model/GlobalDefinitionsIT.java
```diff
@@ -159,6 +158,18 @@ public void testCamelBeanPropertiesSpring() throws IOException, CoreException {
 		assertThat(bean3prop1.getAttribute(GlobalBeanEIP.PROP_NAME)).isEqualTo("description");
 		assertThat(bean3prop1.getAttribute(GlobalBeanEIP.PROP_VALUE)).isEqualTo("#1 account");
 		
+		CamelBean beanWithAllAttributes = (CamelBean)model1.getGlobalDefinitions().get("whateverWithAllFields");
+		checkAllCommonAttributesLoaded(beanWithAllAttributes);
+		assertThat(beanWithAllAttributes.getParameter(GlobalBeanEIP.PROP_FACTORY_BEAN)).isEqualTo("factoryMethod");
+	}
+
+	private void checkAllCommonAttributesLoaded(CamelBean beanWithAllAttributes) {
```
bfitzpat: I think we'll need to split this test into a Spring test and a Blueprint test, but otherwise I think this looks great.

apupier: What do you want to split more?
In this method there is all common attributes in Spring and Blueprint and the different attribute is checked in 2 different test method.
Do you mean that it would be better to have 2 distinct test classes?

bfitzpat: I think because there are distinct differences between Global Bean properties for Spring and Blueprint, it would be good to call that out in the test by providing different methods. It's not as though we need a complete separate class, but it drives me crazy to see the Blueprint and Spring properties mixed.

apupier: I really don't understand why you mean then. There are already 2 methods:
- testCamelBeanPropertiesSpring
- testCamelBeanPropertiesBlueprint

PR #1017, opened 2017-07-20
FUSETOOLS-2460 - Simplify code

apupier:
- it adds also a little feature to have a delay before registering into
model for all properties using a Text

"28 additions and 197 deletions"
@bfitzpat please check that i missed nothing but it should simplify a bit the code for further investigation on the cursor issue


bfitzpat:
Yeah, this doesn't work for me at all. I get all sorts of errors I'm trying to track down when it runs. Validation seems to be completely broken on text fields.

bfitzpat:
First thing is a NPE happening on this line courtesy of the TextParameterPropertyUICreator. 
https://github.com/jbosstools/jbosstools-fuse/blob/master/core/plugins/org.fusesource.ide.camel.validation/src/org/fusesource/ide/camel/validation/model/TextParameterValidator.java#L108

bfitzpat:
If I fix that NPE by putting in a check (camelModelElement.getRouteContainer() != null) it then uses the validator associated with that class instead of the validator I add in the binding. I'm good with the extra level of validation, but only if the binding I actually want gets triggered as well.

bfitzpat:
I'm actually not sure why the validator I create in https://github.com/jbosstools/jbosstools-fuse/blob/master/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/properties/AdvancedBeanPropertiesSection.java#L219 isn't used. It makes no sense to me.

bfitzpat:
I tried creating a PR with some fixes for your PR but I'm sure I screwed it up - anyway... https://github.com/apupier/fuseide/pull/6

bfitzpat:
My PR still doesn't address the issue of the cursor hopping to the front of the field btw - but it makes things work again for the Bean properties.

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/properties/TextParameterPropertyUICreatorWithBeanIDPropertyValidator.java
```diff
@@ -0,0 +1,43 @@
...
+import org.fusesource.ide.camel.model.service.core.catalog.Parameter;
+import org.fusesource.ide.camel.model.service.core.catalog.eips.Eip;
+import org.fusesource.ide.camel.model.service.core.model.AbstractCamelModelElement;
+import org.fusesource.ide.foundation.core.util.CompoundValidator;
+
+class TextParameterPropertyUICreatorWithBeanIDPropertyValidator extends TextParameterPropertyUICreator {
+
+	private final AdvancedBeanPropertiesSection advancedBeanPropertiesSection;
+
+	public TextParameterPropertyUICreatorWithBeanIDPropertyValidator(AdvancedBeanPropertiesSection advancedBeanPropertiesSection, DataBindingContext dbc, IObservableMap<?,?> modelMap, Eip eip,
```
apupier: unfortunately would need a major refactor of this XXXUICreators, i think that we need to live with it for now

PR #1018, opened 2017-07-21
FUSETOOLS-2456 - remove Tracing capabilities as it is not used

apupier:


bfitzpat:
Wouldn't this be helpful to keep even if we're not using it? I can definitely see it being useful down the line at some point

apupier:
> Wouldn't this be helpful to keep even if we're not using it? I can definitely see it being useful down the line at some point

yes it is very useful to provide fine trace for debugging purpose.
But:
- when we will need to provide fine Trace, it will be easy to add those 3 files back
- there are 8 Trace classes in our plugins. Only one is effectively used. Some of the 7 others are here for several years but still not used.
- less code is better code

bfitzpat:
So you say that a) it is very useful but b) you want to get rid of it. It's less than 100 lines that could, possibly be useful at some point in the future. Since it's so small and useful I would vote to leave it in even if we haven't used it in several years. But I'll leave it to you and @lhein to battle this one out.

apupier:
if we keep it, I also want that the issue is reported by Sonar is fixed (or reported as false positive to Sonarsource team)

lhein:
I'd really like to keep it.

apupier:
if you want to keep it, please at least fix the introduced Sonar issues with this class.
(even if I really don't understand the interest of keeping dead code)

PR #1026, opened 2017-07-25
Add missing "/" for directory entry

apupier:


apupier:
> interesting that the extra slash was needed. is it something missing cross-platform? or just on Windows?

to be honest I don't know, Eclipse is reporting this error.
the src without / has been added recently during the move to JBoss Tools

PR #1032, opened 2017-07-25
FUSETOOLS-2473 - Always load a Catalog

apupier:
if version < 2.10.0 --> latest Community version tested in product
(currently 2.19.1)
if 2.10.0 <= version < 2.15.0 --> 2.15.6
if impossible to load --> latest Community version tested in product
(currently 2.19.1)


apupier:
tests have failed with this error:

> java.lang.NoClassDefFoundError: org/apache/camel/catalog/CamelCatalog
> 	at org.fusesource.ide.camel.model.service.internal.CamelServiceCatalogLoadingTest.testCatalogLoad(CamelServiceCatalogLoadingTest.java:71)
> Caused by: java.lang.ClassNotFoundException: org.apache.camel.catalog.CamelCatalog
> 	at org.fusesource.ide.camel.model.service.internal.CamelServiceCatalogLoadingTest.testCatalogLoad(CamelServiceCatalogLoadingTest.java:71)

but i'm nto able to reproduce locally :(

apupier:
> My one comment is that we probably need some way to indicate when the default catalog is loaded vs. one specific to the Camel version of the project. Perhaps a note put into the log would be enough -- just some indicator to explain the behavior the user is seeing

I provided a log in this PR.
I created a JIRA enhancement for a better indicator: https://issues.jboss.org/browse/FUSETOOLS-2475

bfitzpat:
Yeah, I didn't see that error locally either.

apupier:
I modified the PR to have the test as an integration one instead of a unit test as I still have no clue why it is failing only on CI. This is sad but at least it should work.

bfitzpat:
testPR

PR #1036, opened 2017-07-28
Exclude Community version upper than 2.19 as the template is nto working

apupier:
as-is due to Fuse bom version mismatch


apupier:
given the timing and that it was an acknowledged "known issues " for this release, I merge without waiting for the 24 hours

PR #1042, opened 2017-07-31
FUSETOOLS-2132 - Provide more Trace for Validation and Markers

apupier:
to help investigate intermittent failures on CI


bfitzpat:
Why disable validation during project creation? Is the validation interrupting the process?

apupier:
> Why disable validation during project creation? Is the validation interrupting the process?

The validation is not interrupting the process.
The validation is launched several times on the same files as we are modifying the project. We just need to validate when we have finished to modify the project.
It will also avoid to have concurrent validation running which might cause the xsd validation error (perhaps, no proof, just a guess)

bfitzpat:
testPR

lhein:
testPR

lhein:
That PR doesn't build any more.

lhein:
Needs to be rebased correctly.

lhein:
testPR

lhein:
@apupier how about rebase and merge?

apupier:
> @apupier how about rebase and merge?

I will rebase asap but I would wait to merge because the patched code is adding too much technical debt and for no it has not been provided needed information when the test is failing. need to find more trace/exception to log before a useful merge

apupier:
the initial issue has not occurred for several weeks so closing this PR

PR #1043, opened 2017-07-31
FUSETOOLS-2459 - setting scope by default for new global beans

bfitzpat:


apupier:
testPR

apupier:
relaunched the test as it seems to be due to known xsd intermittent issue https://issues.jboss.org/browse/FUSETOOLS-2132

lhein:
testPR

PR #1045, opened 2017-08-01
Fusetools 2397 same camel context id and route

apupier:


core/plugins/org.fusesource.ide.camel.validation/src/org/fusesource/ide/camel/validation/diagram/BasicNodeValidator.java
```diff
@@ -230,8 +209,7 @@ private IMarker createOrReuseMarker(IResource resource, AbstractCamelModelElemen
 	 * @throws CoreException
 	 */
 	private IMarker createMarker(IResource resource, final String message, final int severity, Map<String, Object> attributesForPosition) throws CoreException {
-		IMarker res;
-		res = resource.createMarker(IFuseMarker.MARKER_TYPE);
+		IMarker res = resource.createMarker(IFuseMarker.MARKER_TYPE);
```
bfitzpat: Can we change some of these local variables to use a different name than "res" which confuses it with the IResource a bit?

bfitzpat: Overall though I think this looks great -- centralizes a few things around marker management and validation testing, which is awesome.

PR #1049, opened 2017-08-02
JBIDE-24788 support building & deploying...

nickboldt:
JBIDE-24788 support building & deploying with Maven 3.5

### DIFF
--- a/site/pom.xml
+++ b/site/pom.xml
@@ -1,16 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
-	xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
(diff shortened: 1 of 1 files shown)


site/pom.xml
```diff
@@ -1,16 +1,16 @@
...
-	xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 	<modelVersion>4.0.0</modelVersion>
 	<parent>
 		<groupId>org.fusesource.ide</groupId>
 		<artifactId>project</artifactId>
 		<version>10.0.0-SNAPSHOT</version>
 	</parent>
-
+	<groupId>org.fusesource.ide</groupId>
```
apupier: why do we need to repeat the groupId which is the same than the parent?

nickboldt: it's just cleaner. But you don't actually need this PR.

PR #1052, opened 2017-08-03
FUSETOOLS-1749 - Unify Fuse server adapters categories

apupier:
- put JBoss Fuse Runtime in Red Hat Jboss Middleware category
- put Apache Karaf in Apache category

result:
image
image


apupier:
> am concerned about existing users. what will this do to any local configuration they have for servers and launch configurations?

- information are not stored in the project, so only users upgrading their installation might be impacted
- to be tested if there are some impact, but as it is only categories of the definition and labels that have been updated, I expect that there is no impact

lhein:
I agree. There shouldn't be any impact at all.

bfitzpat:
If there's no impact the user, then +1 :)

PR #1056, opened 2017-08-04
FUSETOOLS-2461 - fixing namespacing of new bean node

bfitzpat:

### DIFF
--- a/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/globalconfiguration/beans/BeanConfigUtil.java
+++ b/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/globalconfiguration/beans/BeanConfigUtil.java
@@ -533,7 +533,8 @@ public class BeanConfigUtil {
 		}
 	}
 
-	private void getNSPrefixes(Node rootNode, String namespaceUri, List<String> prefixes) {
+	private List<String> getNSPrefixes(Node rootNode, String namespaceUri) {
+		List<String> prefixes = new ArrayList<>();
 		NamedNodeMap atts = rootNode.getAttributes();
 		for (int i = 0; i < atts.getLength(); i++) {
 			Node node = atts.item(i);
@@ -548,11 +549,11 @@ public class BeanConfigUtil {
 				}
 			}
 		}
+		return prefixes;
 	}
 	
 	private String getNSPrefixForURI(Node rootNode, String namespaceUri) {
-		List<String> prefixes = new ArrayList<>();
-		getNSPrefixes(rootNode, namespaceUri, prefixes);
+		List<String> prefixes = getNSPrefixes(rootNode, namespaceUri);
 		if (!prefixes.isEmpty()) {
 			return prefixes.get(0);
 		}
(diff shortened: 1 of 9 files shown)


editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/globalconfiguration/beans/BeanConfigUtil.java
```diff
@@ -526,7 +527,8 @@ public void editBeanArgument(Element xmlElement, String type, String value) {
 	}
 	
 	public Element createBeanNode(final CamelFile camelFile, String id, String className) {
-		final String prefixNS = camelFile.getRouteContainer().getXmlNode().getPrefix();
+		// get NS prefix from parent document, not route container node
+		final String prefixNS = camelFile.getRouteContainer().getXmlNode().getOwnerDocument().getPrefix();
```
apupier: we are not providing any templates like that but in case the root document is not bean and is using another prefix we will still have an issue. 
Don't we have a way to find the NS prefix for a specific namespace URI?

bfitzpat: Not that I've seen. I found this code that retrieves ALL the prefixes, but it's still not quite what we want: http://www.java2s.com/Tutorial/Java/0440__XML/GetallprefixesdefineduptotherootforanamespaceURI.htm

apupier: seems that it should be available from https://docs.oracle.com/javase/7/docs/api/javax/xml/namespace/NamespaceContext.html but don't know how to get it.

otherwise, at least, it would be nice to get the "beans" node (in the correct namespace uri) and retrieve the prefix from here instead of the route one which we don't know what it is for sure

and a last thing is that it seems that there is a specific method to create the element with a specific namespace (which might explain the "uncategorized" place when creating a new Bean in global bean editor if it is not used): https://docs.oracle.com/javase/7/docs/api/org/w3c/dom/Document.html#createElementNS(java.lang.String,%20java.lang.String)
but still requires to find the correct prefix

bfitzpat: So you're proposing we do the full search through the entire namespace mess on the root document note (not the route node) to locate either the spring or blueprint URI and then grab the prefix from that? My worry with that is speed.

apupier: I'm proposing to avoid assumptions letting apart some cases.

I don't see a speed issue with getting the "beans" node prefix as it should be just the parent node to where we want to create it and most the time the top node.

bfitzpat: Ok. I'll see what I can do.

bfitzpat: I have some more specific functionality working now. Building some integration tests to ensure that we have a few use cases working.

bfitzpat:
Updated PR -- more specific with spring & blueprint namespace URI/prefix combinations at root and deeper. Also added some tests to ensure that the main cases are handled.

bfitzpat:
retest

bfitzpat:
@apupier Thanks for the suggested improvements. Went ahead an created another commit for the optimization so we only return the first NS/prefix combination for that URI as well.

PR #1058, opened 2017-08-08
FUSETOOLS-2299 - Fix model for choice/enum parameters

apupier:


core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/Parameter.java
```diff
@@ -40,8 +40,8 @@
 	private String required;
 	@JsonProperty
 	private String defaultValue;
-	@JsonProperty
-	private String choice;
+	@JsonProperty("enum")
+	private String[] choice;
```
apupier: please note that the attribute cannot be named "enum" so keep "choice" which was the name used previously

lhein: so the name of the value changed from choice to enum? since when? what about backwards compatibility?

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/Parameter.java
```diff
@@ -86,7 +86,7 @@ public String getGroup() {
 	/**
 	 * @return the choice
 	 */
-	public String getChoice() {
+	public String[] getChoice() {
```
lhein: So its a json array now? wondering about backwards compatibility with older catalog versions.

apupier: Will check for it but I supposed that previously we were transforming the attribute in the tooling.

apupier: the catalog is  the same, it was already using an enum array in 2.15.1
for instance:
"pattern": { "kind": "attribute", "required": "false", "type": "string", "javaType": "org.apache.camel.ExchangePattern", "enum": [ "InOnly", "InOptionalOut", "InOut", "OutIn", "OutOnly", "OutOptionalIn", "RobustInOnly", "RobustOutOnly" ], "deprecated": "false", "description": "Sets the optional ExchangePattern used to invoke this endpoint" }

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/catalog/Parameter.java
apupier: it was already mapped the same way: https://github.com/jbosstools/jbosstools-fuse/blob/9.2.x/tools/ide-catalog-generator/src/main/java/org/fusesource/ide/catalog/generator/model/eip/EIPProperty.java#L31

PR #1060, opened 2017-08-08
FUSETOOLS-2392 - Deactivating mouse wheel on combo requiring long

apupier:
computing

when changing languages or dataformat types, it might require a long
computation time due to download + compile time.
As this feature is making a bad experience to Users (UI freeze), we
choose to deactivate it.


bfitzpat:
Does this fix also handle https://issues.jboss.org/browse/FUSETOOLS-2455 ?

apupier:
> Does this fix also handle https://issues.jboss.org/browse/FUSETOOLS-2455 ?

no

apupier:
> +1 but very interesting bug

this is not a bug from SWT. It is that in Fuse Tooling there is a long running action when selecting a new value in the combo for dataformats and languages because maven dependencies are added automatically.

I deactivated the SWT feature of having the scroll working on the widget hovered to avoid triggering long running operations

lhein:
that PR seems kinda unrelated to FUSETOOLS-2392 though

PR #1061, opened 2017-08-09
FUSETOOLS-2363 - avoid duplicated entries for routes in Common navigator

apupier:


apupier:
> Is there any way we can stress test it a bit more? Perhaps attempt to refresh the editor a few times to get it to re-populate the navigator? The issue for me always cropped up when I was doing other things in the editor.

Unfortunately, I already spent most of the time for this feature trying to have a more complex test without success (implementing the fix was finally quite easy)
I can try again tomorrow with fresh minds, otherwise you can also have a look or we can ask QE to write a UI one.

bfitzpat:
Sounds good. This is probably right up @apodhrad 's alley to come up with a good way to test this.

apupier:
I tried again this morning but in fact there is a limitation on Tree with element of same hashcode so I created a JIRA for a UI test https://issues.jboss.org/browse/FUSETOOLS-2502

PR #1066, opened 2017-08-10
FUSETOOLS-2455 - Provide progress bar when adding data format

apupier:
- moved the Util class to wait for build and refresh job in Camel editor
plugin to be usable in this plugin too


apupier:
here are the videos of the difference, you can notice that now it is waiting a bit more time because it is waiting that workspace build has finished but it avoids to have some freeze although the window already closed:
see for new behavior: https://drive.google.com/open?id=0B1cUBNButvV9Z1I0d2lIVWFTcE0
see for old (current) behavior: https://drive.google.com/open?id=0B1cUBNButvV9OUU1b09Bc3VsbXc

apupier:
> Do we need to employ this same approach with any of the other global wizards if we aren't already?

I have not checked for the others. Might be worthy.

PR #1067, opened 2017-08-11
FUSETOOLS-2504 - don't use bom for camel plugins version management

apupier:
there is a bug in io.fabric8 bom which leads to use always the latest
camel-maven-plugin instead of the expected version


apupier:
the second build triggered (to clean code) failed, i relaunch a new one then :'(

PR #1077, opened 2017-08-24
FUSETOOLS-2376 - categorize palette entries by Catalog information (D…

lhein:
…RAFT)


lhein:
Well...the point is that a lot of catalog items were never displayed and with that PR they finally are. We now need to think about which of them we do want users to see and which better stay hidden.
Also I am not yet fully sure if the categorization makes always sense as is or if we should instead start using the categories in the catalog as palette drawer names.

PR #1079, opened 2017-08-28
FUSETOOLS-2445 - refactor access to stagingrepo preferences

lhein:


lhein:
testPR

lhein:
testPR

core/plugins/org.fusesource.ide.preferences/src/org/fusesource/ide/preferences/Messages.java
```diff
@@ -0,0 +1,39 @@
+/*******************************************************************************
+ * Copyright (c) 2013 Red Hat, Inc.
```
apupier: 2017

core/tests/org.fusesource.ide.preferences.tests.integration/src/main/java/org/fusesource/ide/preferences/tests/PreferencesIntegrationTestsActivator.java
```diff
@@ -0,0 +1,16 @@
+/*******************************************************************************
+ * Copyright (c) 2016 Red Hat, Inc.
```
apupier: 2017

core/tests/pom.xml
```diff
@@ -20,9 +20,10 @@ xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xs
 		<module>org.fusesource.ide.camel.model.service.core.tests.integration</module>
 		<module>org.fusesource.ide.camel.model.service.impl.tests.integration</module>
 		<module>org.fusesource.ide.camel.model.tests.integration</module>
+		<module>org.fusesource.ide.preferences.tests</module>
```
apupier: would be nice if you can place the module entries in the alphabetical order

lhein:
I agree. The settings would be lost. Any suggestions on how to migrate them?

lhein:
rebased with master

lhein:
testPR

apupier:
> The settings would be lost. Any suggestions on how to migrate them?

I have never done that before.
Technically, I imagine that we can do check on a each startup to migrate to the new settings. The downside is that we will have some legacy code just for migration and that the check will occur on each startup (or each first access of the new preferences) i think that there are also p2 touchpoint to do some modification during installation but it might be a bit complicated.
From a Functional point of view, I'm wondering if it is worthy. This functionality is targeting mainly QE, Doc and very advanced users playing with non-official (or not-yet-official) dependencies, so just a warning in release notes might be enough

lhein:
Yep, I tend to agree here. @apodhrad What do u think?

PR #1085, opened 2017-08-31
FUSETOOLS-2516 - activate maven debug output

apupier:
to have more traces on the failing maven launch

### DIFF
--- a/editor/tests/org.fusesource.ide.projecttemplates.tests.integration/src/main/java/org/fusesource/ide/projecttemplates/tests/integration/wizards/FuseIntegrationProjectCreatorRunnableForOSESringBootIT.java
+++ b/editor/tests/org.fusesource.ide.projecttemplates.tests.integration/src/main/java/org/fusesource/ide/projecttemplates/tests/integration/wizards/FuseIntegrationProjectCreatorRunnableForOSESringBootIT.java
@@ -13,6 +13,8 @@ package org.fusesource.ide.projecttemplates.tests.integration.wizards;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import org.eclipse.core.resources.IProject;
+import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
+import org.eclipse.m2e.actions.MavenLaunchConstants;
 import org.fusesource.ide.projecttemplates.adopters.util.CamelDSLType;
 import org.fusesource.ide.projecttemplates.impl.simple.OSESpringBootXMLTemplate;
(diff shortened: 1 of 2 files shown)


editor/tests/org.fusesource.ide.projecttemplates.tests.integration/src/main/java/org/fusesource/ide/projecttemplates/tests/integration/wizards/FuseIntegrationProjectCreatorRunnableForOSESringBootIT.java
```diff
@@ -42,4 +44,9 @@ protected void additionalChecks(IProject project) {
 		assertThat(project.findMember("src/main/java/META-INF/MANIFEST.MF")).as("A bad Manifest has been generated").isNull();
 		assertThat(project.getFile(".settings/fusetooling/Deploy " + project.getName() + " on OpenShift.launch").getLocation().toFile()).exists();
 	}
+	
+	@Override
+	protected void additionalMavenAttributes(ILaunchConfigurationWorkingCopy workingCopy) {
+		workingCopy.setAttribute(MavenLaunchConstants.ATTR_DEBUG_OUTPUT, true);
```
lhein: That will blow up the log file length a lot. Maybe we should consider controlling that via System properties instead of a fixed value so we can enable / disable it without further code changes?

apupier: > That will blow up the log file length a lot.

Yes that's why I activated it only on the specific test requiring it

> Maybe we should consider controlling that via System properties instead of a fixed value so we can enable / disable it without further code changes?

I prefer to have it in the code instead of "somewhere not versioned".

I would like to launch this PR several times first before merging in case we are able to have mor einformation on the "unstable issue" before "polluting" master in any way

apupier: here we go: reproduced at the first try: https://issues.jboss.org/secure/attachment/12424197/MavenLaunchOutput-FuseIntegrationProjectCreatorRunnableForOSESringBootIT-OSESpringBootProject-2.18.1.redhat-000015.txt

apupier:
this PR is more useful as we get the log that we wanted. Will recreate another if we can grab more information

PR #1086, opened 2017-09-01
FUSETOOLS-2405 - amq

bfitzpat:
Luca's fixes worked beautifully. Tested against both Spring & Blueprint on a Fuse 6.3 server and all is well. Also updated the readme.txt to fix the issue Tomas found.


bfitzpat:
Was able to remove some of the version limitations. Good call @apupier !

PR #1087, opened 2017-09-01
FUSETOOLS-2487 - remove dependency which was added by error

apupier:


apupier:
finally this wasn't added by error, no wthat tests are failing,  i remember now that it was needed

PR #1091, opened 2017-09-08
FUSETOOLS-2527 - provide the icon for Karaf and Fuse launch config

apupier:
before:
image

after:
image

notice the icon at the top level tree item in Debug view


lhein:
we might have to change the icons soon but +1

PR #1092, opened 2017-09-08
FUSETOOLS-2042 - hide Global Config tab for Routes element files

lhein:


lhein:
> I think that it would be worthy to havbe a test which is opening a Camel Editor for an xml with and ensuring that there is only 2 tabs and maybe, if not already covered that there is one opening the camel editor with camelcontext tab and that there is 3 tabs

Updated the PR to have tests for that specific case.

PR #1098, opened 2017-09-13
JBIDE-25018 remove discovery...

nickboldt:
JBIDE-25018 remove discovery sites


apupier:
@lhein you told me that you were wondering if it was not used for something else too? Do you remember for what? If yes, if it should be keep or not?

nickboldt:
There is zero reason to include these since they are DISCOVERY, not UPDATE so they're disabled by default. If a user then enables them they'll link to content that's already available from the jbt/devstudio update sites.

PR #1100, opened 2017-09-14
FUSETOOLS-2472 - Return false when not interested in the resource

apupier:
it avoids to scan all other projects, this will improve performance on
large workspaces


apupier:
testPR

PR #1101, opened 2017-09-14
FUSETOOLS-2472 - return true when the connection is valid

apupier:
- the returned value is currently never used so there was no functional
bugs but it would have been the case if tried to implement a feature
relying on it


apupier:
> wow, good catch

we can thanks Sonar https://sonarcloud.io/project/issues?id=fusetools&issues=AVxpFU_fWtfsztvt2Oj7&open=AVxpFU_fWtfsztvt2Oj7

apupier:
testPR

PR #1111, opened 2017-09-22
FUSETOOLS-2550: UI Tests - Preference pages - referencedComposite can't be null

tsedmik:
https://issues.jboss.org/browse/FUSETOOLS-2550

Signed-off-by: Tomáš Sedmík <tsedmik@redhat.com>


apupier:
Shouldn't we remove the contructor ConsolePreferencePage() ? it will force to be in correct state, isn't it?
There are several other XXXPreferencePage which are following the same pattern, should we do the same for them?

tsedmik:
@apupier yes, you are right. I'll look at it again.

tsedmik:
@apupier done

apupier:
seems that you missed FuseServerRuntimePreferencePage

tsedmik:
Good catch. Thank you. _FuseServerRuntimePreferencePage_ is fixed.

apupier:
@apodhrad can you review this PR please as you better know this part of code?

PR #1114, opened 2017-09-25
FUSETOOLS-2547 - Explain why keeping Maven Warning

apupier:


pom.xml
```diff
@@ -83,6 +83,9 @@
 	</modules>
 
 	<repositories>
+		<!-- Start concerned by comment FUSETOOLS-2547 -->
```
lhein: I think concerned by is not the right terminology here. I'd rather put here something like "Start of repository definitions which cause a warning in early build (see FUSETOOLS-2547)"

pom.xml
```diff
@@ -100,6 +103,7 @@
 			<layout>p2</layout>
 			<url>${jbosstools-central-site}</url>
 		</repository>
+		<!-- End concerned by comment FUSETOOLS-2547 -->
```
lhein: I'd replace with "End of repository definitions causing a warning (see FUSETOOLS-2547)"

apupier:
testPR

apupier:
testPR

PR #1118, opened 2017-09-26
FUSETOOLS-2394 - report error validation for same component id and

apupier:
component definition id

in initial issue it was mentioned to add a warning also for cross-components because it is working but not a best practice.
I fear performance issue so maybe better to investigate more for the second part in another PR (perhaps issue too) to decide to decide to do it or not


lhein:
for some reason I get 17 failures when running the validation integration tests. not sure yet if that's related to your PR.

lhein:
I tested on master and that validates fine. So must be a problem in your PR.

lhein:
I still see 2 exceptions when running the integration tests.

bildschirmfoto von 2017-09-27 09-57-26

lhein:
Forget my last comment. I was still using the refactored version of the validator which seem to cause the issue. Your branch works fine.

apupier:
testPR

PR #1120, opened 2017-10-02
FUSETOOLS-2456 - apply Singleton pattern to avoid setting static

apupier:
instances in a constructor

the parent should not be used static either so hard to do better than that but at least it removes the setting of static field in constructors (reported by Sonar as a code smell). I won't fight more on that as it is dead code and my personal point of view is that it should be completely removed.


apupier:
testPR

apupier:
testPR

PR #1121, opened 2017-10-03
FUSETOOLS-2563 - Improve guidance for workaround error

apupier:
image


apupier:
I updated to
image

I kept the "-", I find it clearer but is is something usual in English/American? Or i should really remove them and use the "or"?

apupier:
testPR

apupier:
last failure was due to SonarQube unreachable.
merging as tests were in success

PR #1122, opened 2017-10-03
FUSETOOLS-2562 - log information instead of having a NPE

apupier:
it is not fixing the issue but try to have more information instead of
NPE

### DIFF
--- a/servers/plugins/org.fusesource.ide.server.karaf.core/src/org/fusesource/ide/server/karaf/core/server/subsystems/Karaf2xPublishController.java
+++ b/servers/plugins/org.fusesource.ide.server.karaf.core/src/org/fusesource/ide/server/karaf/core/server/subsystems/Karaf2xPublishController.java
@@ -20,7 +20,6 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
-import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.m2e.core.MavenPlugin;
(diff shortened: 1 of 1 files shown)


servers/plugins/org.fusesource.ide.server.karaf.core/src/org/fusesource/ide/server/karaf/core/server/subsystems/Karaf2xPublishController.java
```diff
@@ -263,8 +263,13 @@ private IPath getTempBundlePath(IModule[] module) {
     private String getModuleVersion(IModule module) throws CoreException {
         IProgressMonitor monitor = new NullProgressMonitor();
         IProject project = module.getProject();
-        IMavenProjectFacade facade = MavenPlugin.getMavenProjectRegistry().create(project.getFile(IMavenConstants.POM_FILE_NAME), true, monitor);
-        MavenProject mavenProject = facade.getMavenProject(monitor);
-        return mavenProject.getVersion();
+        if (project == null) {
+        	Activator.getLogger().warning("Cannot determine project associated with module: " + module.getName());
+        	return "";
```
lhein: I think we should not add another possibility for the return value here. In the calling method we catch CoreException and assign a module version in that case already. I would suggest to throw a CoreException instead of returning an empty string.

apupier: I was wondering if I should throw an exception or not but I'm not sure that it would cause trouble if the version is not found.
I think that we will need to check that to know if it is better to throw an exception or returning an empty value.

apupier: Thanks for pointing the code around the method, I decided to group all the information to retrieve module version in the same method, it avoids to have exceptions thrown and catch through different methods.

apupier:
testPR

apupier:
the fuse CI is green (the only working one for now)

PR #1123, opened 2017-10-04
JBDS-4574 - Don't override runtime name with folder name

apupier:
it helps to keep consistent naming, especially when using Dev Suite installer


apupier:
I'm not waiting for the 24h as we are only 2 dev this week, so "all" dev had put a +1 ^^

PR #1127, opened 2017-10-06
Update readme

apupier:
- update link to forum which has moved
- add information about Mattermost
- add information about Twitter and developers.redhat blogposts


apupier:
ignoring work on new CI with failing build due to infra

PR #1129, opened 2017-10-10
FUSETOOLS-2574 - provide indication to use jdk instead of jre

apupier:


apupier:
main CI is green, other failure are due to not ready CI

PR #1152, opened 2017-10-27
remove useless assignments

apupier:
as the variable already holds the assigned value along all execution
paths.


apupier:
testPR

PR #1156, opened 2017-10-31
FUSETOOLS-2536 - provide progress bar when switching camel version

apupier:
compare old version with "not responding" (where i even lost my window for few seconds because it went in the background...)
https://drive.google.com/open?id=0B1cUBNButvV9Mi1UdWpTZW92UUk
with new version:
https://drive.google.com/open?id=0B1cUBNButvV9Yi1nUkhrbzRzOGM


bfitzpat:
Definitely an improvement!

PR #1161, opened 2017-11-06
FUSETOOLS-2451 - Provide a CamelContext EIP from Catalog for 2.18-

apupier:
version

it was missing from Camel catalog by default, it was provided since
Camel 2.18


lhein:
And you made sure that from the 1st Catalog provided up to the latest 2.17.x catalog the properties and defaults did not change a single time?

apupier:
> And you made sure that from the 1st Catalog provided up to the latest 2.17.x catalog the properties and defaults did not change a single time? It was a long way from 2.15.x to 2.18.

I was checking that. Effectively, there are several differences.
Not sure how to generate the json file automatically for old versions, there are several commits to implement that.

apupier:
I updated the PR to contain a specific Camel Context model per 2.y version from 2.10.z to 2.17.z
I don't think that it is worthy to spend time on the various z variations to check if something has changed or not.

PR #1162, opened 2017-11-08
FUSETOOLS-2584: Allow to specify own configuration for camel2bom.properties

apodhrad:


apodhrad:
If we wanted to add some tests, we would have to change the logic of CamelCatalogUtils. Currently, all properties are resolved at the moment when the class is used for the first time. Then, it is not possible to change the properties during runtime.

apodhrad:
Btw the change is trivial so there's no need for tests ;-)

PR #1172, opened 2017-11-24
Fusetools 2195 select runtime camel version filter templates

apupier:
@lhein @bfitzpat What do you think to this way to provide different templates depending on camel version chosen in previous page?

- I think it will be easy to extend to filter base don Runtime too after.
- I think that it will greatly help with the Fuse 7 stuff, we will be able to provide a new Template completely different
- it allows to filter out the OpenShift template for Camel version < .18 which are not working for sure.

### DIFF
--- a/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/AbstractProjectTemplate.java
+++ b/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/AbstractProjectTemplate.java
@@ -84,4 +84,8 @@ public abstract class AbstractProjectTemplate {
 	 * @return the creator
 	 */
 	public abstract TemplateCreatorSupport getCreator(NewProjectMetaData projectMetaData);
+
+	public boolean isCompatible(String camelVersion) {
+		return true;
+	}
 }
--- a/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/impl/simple/AMQTemplate.java
+++ b/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/impl/simple/AMQTemplate.java
@@ -13,6 +13,7 @@ package org.fusesource.ide.projecttemplates.impl.simple;
 import java.io.IOException;
 import java.io.InputStream;
 
+import org.apache.maven.artifact.versioning.ComparableVersion;
 import org.fusesource.ide.projecttemplates.adopters.AbstractProjectTemplate;
 import org.fusesource.ide.projecttemplates.adopters.configurators.MavenTemplateConfigurator;
 import org.fusesource.ide.projecttemplates.adopters.configurators.TemplateConfiguratorSupport;
@@ -25,6 +26,8 @@ import org.fusesource.ide.projecttemplates.util.NewProjectMetaData;
  * @author lhein
  */
 public class AMQTemplate extends AbstractProjectTemplate {
+	
+	private static ComparableVersion maximalCompatibleCamelVersion = new ComparableVersion("2.20.0");
 
 	@Override
 	public boolean supportsDSL(CamelDSLType type) {
@@ -34,6 +37,11 @@ public class AMQTemplate extends AbstractProjectTemplate {
 		case JAVA:		return false;
 		default:		return false;
 		}	
+	}
+	
+	@Override
+	public boolean isCompatible(String camelVersion) {
+		return new ComparableVersion(camelVersion).compareTo(maximalCompatibleCamelVersion) < 0;
 	}
 	
 	@Override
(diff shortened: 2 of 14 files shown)


apupier:
example with version < 2.18 filtering OpenShift template:
image
image

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/FuseIntegrationProjectWizard.java
```diff
@@ -41,25 +41,11 @@ public FuseIntegrationProjectWizard() {
...
 	@Override
 	public void init(IWorkbench workbench, IStructuredSelection selection) {
 		this.selection = selection;
 	}
 
-	/* (non-Javadoc)
-	 * @see org.eclipse.jface.wizard.Wizard#needsProgressMonitor()
-	 */
-	@Override
-	public boolean needsProgressMonitor() {
```
apupier: not needed, l.41 is already doing the job

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/FuseIntegrationProjectWizardRuntimeAndCamelPage.java
```diff
@@ -175,7 +170,10 @@ public void keyReleased(KeyEvent e) {
 			}
 		});
 
-		camelVersionCombo.addModifyListener(e -> validate());
+		camelVersionCombo.addModifyListener(e -> {
+			validate();
+			((FuseIntegrationProjectWizardTemplatePage)getWizard().getPage(Messages.newProjectWizardTemplatePageName)).refresh(camelVersionCombo.getText());
```
apupier: I'm not a big fan of that but at least it avoids coupling the wizard pages too much with the filter
I think that the best would be to have an "Object Model" for the whole wizard which is shared between all pages and put listener on it but seems a bit complicated to implement.
What do you think?
Other suggestions?

apupier:
testPR

bfitzpat:
I like the idea of filtering of templates, but it might be good to provide some sort of feedback to the user that not all the available templates are visible for that particular one, along with perhaps a way to override the filtering if they want to see others -- with the explicit warning that it may not work with the version they have selected.

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/impl/simple/AMQTemplate.java
```diff
@@ -36,6 +39,11 @@ public boolean supportsDSL(CamelDSLType type) {
 		}	
 	}
 	
+	@Override
+	public boolean isCompatible(String camelVersion) {
+		return new ComparableVersion(camelVersion).compareTo(maximalCompatibleCamelVersion) < 0;
```
bfitzpat: So is there at maximum only one comparable version for a template? What if it doesn't change between versions?

apupier: only one version is compared at a time which is the version selected by user in the previous Wizard Page.

apupier:
> it might be good to provide some sort of feedback to the user that not all the available templates are visible for that particular one,

I provided a text with a little information icon below the Template list:
image
image


> along with perhaps a way to override the filtering if they want to see others -- with the explicit warning that it may not work with the version they have selected.
I think that I would prefer to create another JIRA issue for that as it requires more work. Several things to take care:
- implement refresh of the list based on the checkbox
- ensure all templates have different names if we have different templates for same usage but for different version (I think that it will be the case for AMQ)
- would be nice to have feedback which ones "should have been" filtered even if they appear in the list

bfitzpat:
The message is great Aurelien. And sounds good to move the additional work into a separate JIRA.

apupier:
I created https://issues.jboss.org/browse/FUSETOOLS-2681

apupier:
testPR

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/adopters/AbstractProjectTemplate.java
```diff
@@ -84,4 +84,8 @@ private void refreshProjectSync(IProject project, IProgressMonitor monitor) thro
 	 * @return the creator
 	 */
 	public abstract TemplateCreatorSupport getCreator(NewProjectMetaData projectMetaData);
+
+	public boolean isCompatible(String camelVersion) {
```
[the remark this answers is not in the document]
apupier: it is used in subclasses

editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/filter/CompatibleCamelVersionFilter.java
```diff
@@ -0,0 +1,53 @@
...
+			return ((TemplateItem) element).isCompatible(camelVersion);
+		} else if (element instanceof CategoryItem) {
+			return hasCompatibleChildren((CategoryItem)element);
+		}
+		return true;
+	}
+
+	private boolean hasCompatibleChildren(CategoryItem category) {
+		boolean hasCompatibleTemplate = category.getTemplates().stream()
+				.anyMatch(template -> template.isCompatible(camelVersion));
```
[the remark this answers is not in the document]
apupier: hum sonar is reporting a bad thing, it is already anyMatch as we can see

PR #1175, opened 2017-11-28
FUSETOOLS-2674 - comment springboot-devtool as it seems to cause issue

apupier:
with local debugger


apupier:
not waiting 24 hours as it i minimal change and blocking several other PRs

PR #1176, opened 2017-11-29
FUSETOOLS-2683: Update RedDeer launch file

apodhrad:


apodhrad:
I think it is not necessary to wait 24h for such small and easy change

apupier:
> I think it is not necessary to wait 24h for such small and easy change

the 24h rule is to allow everybody to have a look to the PR so better to respect it unless it is blocking other work

PR #1177, opened 2017-11-30
FUSETOOLS-2675: Implement a requirement for Camel examples

apodhrad:

### DIFF
--- a/uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/requirement/CamelExampleRunner.java
+++ b/uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/requirement/CamelExampleRunner.java
@@ -1,0 +1,177 @@
+/*******************************************************************************
+ * Copyright (c) 2017 Red Hat, Inc. 
+ * Distributed under license by Red Hat, Inc. All rights reserved. 
+ * This program is made available under the terms of the 
+ * Eclipse Public License v1.0 which accompanies this distribution, 
+ * and is available at http://www.eclipse.org/legal/epl-v10.html 
+ * 
+ * Contributors: 
+ * Red Hat, Inc. - initial API and implementation
+ ******************************************************************************/
+package org.jboss.tools.fuse.reddeer.requirement;
+
+/**
+ * Runner for managing local Camel route.
+ * 
+ * @author Andrej Podhradsky (apodhrad@redhat.com)
+ * 
+ */
+import java.io.File;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+import org.eclipse.reddeer.common.condition.WaitCondition;
+import org.eclipse.reddeer.common.wait.TimePeriod;
+import org.eclipse.reddeer.common.wait.WaitUntil;
+import org.jboss.tools.fuse.reddeer.utils.JarRunner;
+
+public class CamelExampleRunner extends JarRunner {
+
+	public static final Pattern IS_STARTED_PATTERN = Pattern.compile(".*Apache Camel .* started in .*");
+	public static final Pattern IS_SUSPENDED_PATTERN = Pattern.compile(".*Apache Camel .* is suspended in .*");
+	public static final Pattern IS_RESUMED_PATTERN = Pattern.compile(".*Apache Camel .* resumed in .*");
+
+	private CamelExampleOutput output;
+
+	public CamelExampleRunner(File jarFile) {
+		this(jarFile, false);
+	}
+
+	public CamelExampleRunner(File jarFile, File jolokiaJarFile) {
+		this(jarFile, false);
+	}
+
+	public CamelExampleRunner(File jarFile, boolean useJolokia) {
+		super(jarFile);
+		output = new CamelExampleOutput();
+		setOutput(output);
+
+		registerPattern(IS_STARTED_PATTERN);
+		registerPattern(IS_SUSPENDED_PATTERN);
+		registerPattern(IS_RESUMED_PATTERN);
+	}
+
+	@Override
+	public void run(String... args) {
+		super.run(args);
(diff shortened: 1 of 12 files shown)


uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/utils/JarRunner.java
```diff
@@ -0,0 +1,135 @@
...
+
+	protected List<String> getCommand(String... args) {
+		List<String> command = new ArrayList<String>();
+		command.add(System.getProperty("java.home") + "/bin/java");
+		if (javaAgent != null) {
+			command.add("-javaagent:" + javaAgent);
+		}
+		systemProperties.forEach((key, value) -> command.add("-D" + key + "=" + value));
+		command.add("-jar");
+		command.add(jarFile.getAbsolutePath());
```
apupier: in case there spaces in path on Windows, I think that you need to put quotes around it

uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/utils/JarRunner.java
```diff
@@ -0,0 +1,135 @@
...
+		try {
+			input.join();
+		} catch (InterruptedException ie) {
+			throw new JarRunnerException("Cannot stop the process", ie);
+		}
+		process.destroy();
+	}
+
+	protected List<String> getCommand(String... args) {
+		List<String> command = new ArrayList<String>();
```
apupier: can be written using diamond operator: List<String> command = new ArrayList<>();

uitests/tests/org.jboss.tools.fuse.ui.bot.tests/src/org/jboss/tools/fuse/ui/bot/tests/suite/WithoutServerTests.java
```diff
@@ -40,7 +39,6 @@
 	DebuggerTest.class,
 	DownloadServerTest.class,
 	FuseProjectTest.class,
-	JMXNavigatorTest.class,
```
apupier: just to better understand: the test is removed from Smoke and WithoutServer test suite.
What is the reason, it takes too much time?
As soon as there is a requirement we can't have them in these test suites?

apodhrad: WithoutServerTests is executed without any requirements. Since JMXNavigatorTest now requires a requirement for a Camel example, I had to remove the test fro the test suite.

In the future we plan to rewrite and remove some test suites according to the integration with Polarion.

uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/utils/JarRunner.java
apodhrad: good point!

apodhrad: we also need quotes when setting -javaagent

apodhrad: So, it seems that we cannot use the quotes. Notice that the command is a list of strings and the ProcessBuilder processes each item no matter if there is a space or not.

uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/requirement/CamelExampleRunner.java
```diff
@@ -0,0 +1,174 @@
...
+				if (pattern.matcher(csq).matches()) {
+					patternFound.add(pattern);
+				}
+			}
+			return this;
+		}
+
+		@Override
+		public Appendable append(CharSequence csq, int start, int end) throws IOException {
+			return append(csq);
```
tsedmik: It is a wrong implementation

> Appends a subsequence of the specified character sequence to this Appendable.
An invocation of this method of the form out.append(csq, start, end) when csq is not null, behaves in exactly the same way as the invocation `out.append(csq.subSequence(start, end))`

uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/requirement/CamelExampleRunner.java
```diff
@@ -0,0 +1,174 @@
...
+			StringBuffer buffer = new StringBuffer();
+			getLines().forEach(line -> buffer.append(line + System.getProperty("line.separator")));
+			return buffer.toString();
+		}
+
+		@Override
+		public Appendable append(CharSequence csq) throws IOException {
+			lines.add(csq.toString());
+			if (lines.size() > LIMIT) {
+				lines.remove(0);
```
tsedmik: What about _patternFound_ set? It could contain a pattern which is no longer present in _lines_.

apodhrad: That's ok. The point here is that we do not want to keep the full history in memory but we want to keep "checkpoints" such as the route has been started etc.

uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/requirement/CamelExampleRunner.java
apodhrad: The implementation was designed just for our concrete purpose, that's why it is private.

apodhrad: but I can fix it, of course ;-)

PR #1181, opened 2017-12-04
FUSETOOLS-2651: Implement a requirement for Jolokia connections

apodhrad:


apodhrad:
Merged together with #1182

PR #1183, opened 2017-12-05
upgraded TP's IU versions

lhein:


lhein:
not yet ready to review. We are still missing some deps of as.tests.core feature

apodhrad:
@lhein what about using version "0.0.0" in the future? Do we really need to specify the version?

PR #1207, opened 2018-01-04
FUSETOOLS-2401 - Revert class introducing regression for cancel not

apupier:
working

regression introduced by:
https://github.com/jbosstools/jbosstools-fuse/commit/bf39873f2c44a50e4ab8f59f8401d78235caf477

# Pull Request Checklist

## General

## Function

- [ ] Does the project still build fine locally?
- [ ] Does your modification work?
- [ ] Are other parts that use the same component still working fine?

## Code Style

- [ ] Are methods concise, not too long?
- [ ] Are catch blocks catching precise Exceptions only (no catch all)?
- [ ] Have you set the correct year in the headers of new files?

## Tests

- [ ] Are there unit-tests?

### DIFF
--- a/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/util/CamelVersionChecker.java
+++ b/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/util/CamelVersionChecker.java
@@ -13,11 +13,8 @@ package org.fusesource.ide.projecttemplates.util;
 import java.lang.reflect.InvocationTargetException;
 
 import org.eclipse.core.runtime.IProgressMonitor;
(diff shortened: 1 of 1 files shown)


apupier:
@tsedmik do you have automated test checking for cancellation when there are unknown Camel version? (As you caught it, to know if it was a manual test or an automated one and see how much I should try to write one)

tsedmik:
There is no automated test for this scenario yet. I raised an issue to create it - see https://issues.jboss.org/browse/FUSETOOLS-2735.

apupier:
I managed to be in previous state and in fact we can't cancel 2 times a verify (unless the Thread launched in background has finished)
I need to check that it was really already behaving like that before.

apupier:
I will create another PR moving forward instead of reverting as the issue was more common only when there was outage fo several maven repositories

PR #1234, opened 2018-01-17
FUSETOOLS-2668: Support new Dozer 6 xsd coming with Camel 2.20

jpav:


targetplatform/JBTIS Oxygen Mac OS.launch
```diff
@@ -20,8 +20,8 @@
 <stringAttribute key="org.eclipse.jdt.launching.VM_ARGUMENTS" value="-Xms256m &#10;-Xmx1024m"/>
```
lhein: that file does not really belong to this PR

servers/plugins/org.fusesource.ide.server.karaf.core/src/org/fusesource/ide/server/karaf/core/util/KarafUtils.java
```diff
@@ -326,10 +326,12 @@ public static String getBundleSymbolicName(IModule module) throws CoreException
 	
 	public static String stripParametersFromSymbolicName(String symbolicName) {
 		String resVal = symbolicName;
-		// sometimes parameters are added to the symbolic name - we should ignore them
-		int paramIdx = symbolicName.indexOf(';');
-		if (paramIdx != -1) {
-			resVal = symbolicName.substring(0, paramIdx);
+		if (symbolicName != null) {
```
apupier: might worth to rebase on master to avoid seeing this modification in the Pull Request

transformation/plugins/org.jboss.tools.fuse.transformation.core/src/org/jboss/tools/fuse/transformation/core/dozer/DozerMapperConfiguration.java
```diff
@@ -50,7 +50,7 @@
     private static final String CUSTOM_MAPPER_ID = "_customMapping"; //$NON-NLS-1$
     private static final String EXPRESSION_MAPPER_ID = "_expressionMapping"; //$NON-NLS-1$
     private static final String DOZER_SCHEMA_LOC =
-            "http://dozer.sourceforge.net http://dozer.sourceforge.net/schema/beanmapping.xsd"; //$NON-NLS-1$
+            "http://dozermapper.github.io/schema/bean-mapping http://dozermapper.github.io/schema/bean-mapping.xsd"; //$NON-NLS-1$
```
apupier: the exact url is https://dozermapper.github.io/schema/bean-mapping.xsd
Eclipse is not able to follow redirection and display an error

apupier:
the commit used is an amendment of another commit and not a new one.
--> need to reseparate them
- the Jira number and commit message is then not correct

apupier:
I rebased, modified the wrong commit message and forced push on the branch including a fix for Namespace url.

apupier:
I'm wondering if it makes sense to change the namespace value in the documentation as-is. It should better to have a distinction between 6.3 and 7.0. On the other hand, the current implementation is setting the new namespace even for 6.3.

apupier:
I provided a new commit to set back the warning message with updated message when switching from older Camel Version to new Camel Version

apupier:
with the current state @tsedmik has tested with the current "complex Data Transformation" available in QE tests the "Run as local Camel Context" works for 2.15.1 and 2.20.1
So for the most complex automated use case that we have, using the new namespace for dozer with the old version of Dozer is working.
We have found no place mentioning that it is something that can work. Are we ready to bet on it?

apupier:
I reported some other checks based on Dozer migration guid: https://issues.jboss.org/browse/FUSETOOLS-2668?focusedCommentId=13520134&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13520134 I don't think that these others findings can block this PR from being merged

apupier:
There are 2 points to acknowledge:
- bet on using new namespace with older Camel project see https://github.com/jbosstools/jbosstools-fuse/pull/1234#issuecomment-358314684
- Is having the Data Transformation editor for Tech preview 3 outweight the fact that for users upgrading Fuse Tooling and still using Fuse 6.3, the data transformation editor will fail to open without explanation. For which the workaround will be to put the new namespace manually in the xml file?

apupier:
>There are 2 points to acknowledge:  
> - bet on using new namespace with older Camel project see #1234 (comment)

Risk seems low, management ok to take it.

> - Is having the Data Transformation editor for Tech preview 3 outweight the fact that for users upgrading Fuse Tooling and still using Fuse 6.3, the data transformation editor will fail to open without explanation. For which the workaround will be to put the new namespace manually in the xml file?

Management said that a mention in release notes will make it ok

PR #1238, opened 2018-01-17
load versions for ignite extension wizard dynamically from file in this repo

lhein:
Signed-off-by: Lars Heinemann <lhein.smx@gmail.com>

# Pull Request Checklist

After this checklist is all checked or PR provides explanations for possible pass-through, please put the label "Ready for review" on the PR.

## General

- [ ] Did you use the Jira Issue number in the commit comments?
- [x] Did you set meaningful commit comments on each commit?
- [x] Did you sign off all commits for this PR? (git commit -s -m "jira issue number - your commit comment")

## Functional

- [ ] Did the CI job report a successful build?
- [x] Is the non-happy path working, too?

## Maintainability

- [ ] Are all Sonar reported issues fixed or can they be ignored?
- [x] Is there no duplicated code?
- [x] Are method-/class-/variable-names meaningful?

## Tests

- [x] Are there unit-tests?
- [ ] Are there integration tests (or at least a jira to tackle these)?

## Legal

- [x] Have you used the correct file header copyright comment?
- [x] Have you used the correct year in the headers of new files?

PR #1248, opened 2018-01-22
FUSETOOLS-2772 - upgrade ignite version to 1.2.3

lhein:

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?

## Legal

### DIFF
--- a/configuration/ignite.properties
+++ b/configuration/ignite.properties
@@ -1,3 +1,3 @@
 spring.boot.version=1.5.8.RELEASE
 camel.version=2.20.1
-syndesis.version=1.2-SNAPSHOT
+syndesis.version=1.2.3
--- a/syndesis/plugins/org.fusesource.ide.syndesis.extension.core/src/org/fusesource/ide/syndesis/extensions/core/util/SyndesisExtensionsUtil.java
+++ b/syndesis/plugins/org.fusesource.ide.syndesis.extension.core/src/org/fusesource/ide/syndesis/extensions/core/util/SyndesisExtensionsUtil.java
@@ -73,7 +73,7 @@ public class SyndesisExtensionsUtil {
 				
 		model.setCamelVersion(vMapping.getProperty(KEY_CAMEL_VERSION, "2.20.1"));
 		model.setSpringBootVersion(vMapping.getProperty(KEY_SPRING_BOOT_VERSION, "1.5.8.RELEASE"));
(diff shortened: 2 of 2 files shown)


lhein:
testPR

lhein:
@apupier here is the PR to upgrade to 1.2.3 ignite version

apupier:
still same question than here https://github.com/jbosstools/jbosstools-fuse/pull/1238#discussion_r162873185
Where can we found the list of versions for Tech preview 3?
there is also 1.2.6 version for instance.

lhein:
@rhuss told me that 1.2.3 is the version for TP3. I have no idea if the maintain an official document about versions.

apupier:
the list is available here https://docs.google.com/spreadsheets/d/1ohR6poCaYmQ7Ga3OWgSd189sT8MlXexDCZbmb_SY994/edit#gid=0

so it means that the version planned t be used is the upstream version and not the productized version. Is it fine?

lhein:
Interesting...it is the first time I hear from a productized version.

lhein:
ok, seems we need to update the PR to use version 1.2.3.fuse-000001 instead. Will do that asap.

lhein:
done

apupier:
I had in mind to put the Tech preview 3 version. Having the .fuse version will require users to have access to staging repositories. Are they public for Fuse 7 already? Are they configured by default?

apupier:
I think that the PR needs to be rebased to be succesful Only qualifier changed for (org.fusesource.ide.syndesis.extension.core/10.2.0.v20180129-1346). Expected to have bigger x.y.z than what is available in baseline (10.2.0.v20180117-2222)

lhein:
rebased

apupier:
the failure seems to be due tto the fact that the test is relying on the online version available on master

lhein:
of course it does. so I guess we can't expect this PR to build successfully this way.

apupier:
All other tests are passing.
I tested manually through the wizard that it is working with 1.2.3.
I launched the test with forcing usage of 1.2.3 version fro Syndesis and it is working.

This fix should fix the last known master failure.

This issue is currently impacting end users on Fuse Tooling 10.2.0

I started to look to modify the code but with minimal changes, it will mean stopping to test the behavior that users will have (as we won't rely on the online version that they use)

Given that I think that we should merge this PR.

PR #1259, opened 2018-01-31
FUSETOOLS-2763: Provide automatic migration of dozer file when switch…

jpav:
…ing camel version from < 2.20 to >= 2.20

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?

## Legal


lhein:
U need to sign your commits with git commit -s -m "some comment"

apupier:
@jpav have you checked my feedback? What do you think? Do you know when you will have time to discuss/address it?

jpav:
I've addressed most of the concerns in the review, but the ones related to going back and forth between Camel versions I did not.  The reason is the namespace URL is incorporated in the generated models within the project, and there isn't an easy way to handle both scenarios at the same time without a ton more code, which doesn't make sense to me since this is just for migration to a version that we shown handles old transformations as well as new when the URL is updated. In order to allow for the preservation of old projects, I've added a confirmation dialog to allow the user to prevent the conversion if they like, for which the "cancel" option will prevent the transformation editor from opening.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/l10n/messages.properties
```diff
@@ -186,11 +186,13 @@ TransformationEditor_helpTextTarget=Create a new mapping below by dragging a var
 TransformationEditor_invalidPomConfiguration=Unable to recognize project's POM configuration.
 TransformationEditor_invalidTransformationFile=Invalid transformation file: Must be a Dozer configuration file.
 TransformationEditor_messageDialogConfirmation=Are you sure?\n\nAll incomplete mappings will be lost when the editor is closed.
+TransformationEditor_migrationDialogConfirmation=The transformation file must be converted to support the current Camel version.\n\nAre you sure you want to continue?
```
apupier: I think that we should provide more details to user and it is not true that it is "support the current Camel version", it is for the Data Transformation editor to continue support Camel 2.20- versions:
- what is modified?
- what are the consequences if "want to continue"? if "does not want to continue"?

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/l10n/messages.properties
```diff
@@ -186,11 +186,13 @@ TransformationEditor_helpTextTarget=Create a new mapping below by dragging a var
...
 TransformationEditor_mustBeOpenedViaCamelEditor=The transformation editor must be opened via the Camel editor.
 TransformationEditor_tooltipHideSourceVariableViewers=Hide the source/variables viewers
 TransformationEditor_tooltipHideTargetViewers=Hide the target viewer
 TransformationEditor_tooltipShowSourceVariableViewers=Show the source/variables viewers
 TransformationEditor_tooltipShowTargetViewers=Show the target viewer
+TransformationEditor_unableToOpenIncompatibleTransformationFile=Unable to open incompatible transformation file
```
apupier: Why is it incompatible? It would be nice to precise to user.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/TransformationEditor.java
```diff
@@ -349,8 +356,38 @@ public void init(IEditorSite site,
 
         IFile configFile = ((FileEditorInput)getEditorInput()).getFile();
         IJavaProject javaProject = JavaCore.create(configFile.getProject());
-        try {
-            loader = (URLClassLoader)JavaUtil.getProjectClassLoader(javaProject, getClass().getClassLoader());
+        IProgressMonitor monitor = new NullProgressMonitor();
+        try {loader = (URLClassLoader)JavaUtil.getProjectClassLoader(javaProject, getClass().getClassLoader());
```
apupier: what do you think to extract the code responsible to do the migration in a specific method (or even class)? it will improve readability by avoiding a method of 74 lines and will more easily allow to reuse thee code for improving the "Change camel version"
We can also imagine more easily a unit test checking the replacement of the values in the file.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/l10n/messages.properties
apupier: Proposal:
"The transformation file needs to be updated to continue to use Data Transformation editor. A newer Dozer namespace corresponding to Dozer 6.1+/Camel 2.20+ is required. Updated file will continue to work with previous Camel/Dozer version.

Do you want to proceed to automatic update to use the Data Transformation editor?"

for the "Updated file will continue to work with previous Camel/Dozer version.", i'm not sure if we can write it as it is only "based on our current testing":
- it might change in the future (if the Dozer xsd is modified)
- Dozer is saying nothing about this backward-compatibility

apupier:
> I've addressed most of the concerns in the review, but the ones related to going back and forth between Camel versions I did not. The reason is the namespace URL is incorporated in the generated models within the project, and there isn't an easy way to handle both scenarios at the same time without a ton more code, which doesn't make sense to me since this is just for migration to a version that we shown handles old transformations as well as new when the URL is updated. In order to allow for the preservation of old projects, I've added a confirmation dialog to allow the user to prevent the conversion if they like, for which the "cancel" option will prevent the transformation editor from opening.

Can you create tasks in JIRA for the behavior not covered by this PR please? so that we can be sure about the functional scope that this PR is covering and not covering. And be sure to tackle them in a second iteration in near future.

jpav:
Not sure what you mean by the additional JIRA tasks to cover behavior not covered in this PR.  What is not covered? Do you mean the toggling back and forth depending on current Camel version? Or the tests maybe?

For the first bullet, I didn't close the editor because this code is all executed when the editor is opened, and I didn't know of any way to abort the opening. I think I can by using a spawned thread to close it, but do you know of a better way?

I'll make the change for your second bullet above.

Regarding the third bullet, I need to discuss a couple of things on-line with you about this.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/l10n/messages.properties
jpav: I think you're wanting to give too much detail to the user.  This file is supposed to be entirely handled under the covers by the tooling, and should NEVER be modified by the user.  I don't think the user should be concerned about the details of the incompatibility, but I could update the message to say it's related to not supporting Camel 2.2.0? WDYT?

apupier: > This file is supposed to be entirely handled under the covers by the tooling, and should NEVER be modified by the user.

I think that it is possible for a user to modify this file as it is a normal Dozer file.
It might be even almost normal for Camel users which were not using the Fuse Tooling before. (I still hope that we are converting users :-) )

> I could update the message to say it's related to not supporting Camel 2.2.0? WDYT?

yes, would be nice to precise that the file need to be compatible with Dozer version used by Camel 2.20+

jpav: it is definitely NOT the case that the Dozer file should ever be modified by the user. Despite being a Dozer file, it has always been the goal to hide this file from the user, especially since we add our own custom annotations/modifications to the file that Dozer does not inherently understand. Unless you feel strongly otherwise, I'll update the messages as you suggest without diving into particulars about the Dozer file contents.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/l10n/messages.properties
jpav: Regarding your last point, I think we need to rely on the fact that Dozer isn't mentioning anything about the upgrade that breaks backward-compatibility.  I don't see any other alternative than relying on whatever testing we have performed/will perform to ensure this. Future issues we'll deal with when we see future updates to Dozer; there's always the possibility that changes in the future will break things for any software library.

jpav:
Ugh! I'm still battling getting a PDE test to do what I want, and I feel I'm spending way too much time on this, so for now, I'm just going to create the JIRA task you mentioned to create good tests for the editor.

apupier:
Unless I missed something, the current version is not working. I created a PR https://github.com/jpav/jbosstools-fuse/pull/3 with test emphasizing it.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/l10n/messages.properties
apupier: > we add our own custom annotations/modifications to the file that Dozer does not inherently understand.

it is a Camel Dozer stuff which are added? not Fuse Tooling specific?

apupier: > it is definitely NOT the case that the Dozer file should ever be modified by the user

How Camel Dozer users are doing when not using Fuse Tooling? My point was to be able to welcome them and that this point will not block them to adopt Fuse Tooling.

jpav: The changes we add the to file are indeed specific to the tooling. We've modified the runtime to be able to process these changes, but there is no facility provided to add/deal with these changes outside of the tooling. It's been like this from day one.

jpav: The goal with Dozer has always been to provide a transformation engine that could, if we chose, be replaced under the covers someday with another engine with minimal impact to the user. Not sure if we really accomplished that, but that was the goal.

jpav:
Ah, duh, that makes much more sense.  I was stuck on getting a PDE to work. I'll incorporate your test and figure out what's wrong. Thanks.

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/l10n/Messages.java
```diff
@@ -207,11 +207,13 @@
 	public static String TransformationEditor_invalidPomConfiguration;
 	public static String TransformationEditor_invalidTransformationFile;
 	public static String TransformationEditor_messageDialogConfirmation;
+    public static String TransformationEditor_migrationDialogConfirmation;
```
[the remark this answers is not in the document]
lhein: some issue with indentation and sonar reports an issue (see above)

transformation/plugins/org.jboss.tools.fuse.transformation.editor/src/main/java/org/jboss/tools/fuse/transformation/editor/internal/l10n/Messages.java
```diff
@@ -207,11 +207,13 @@
...
 	public static String TransformationEditor_mustBeOpenedViaCamelEditor;
 	public static String TransformationEditor_tooltipHideSourceVariableViewers;
 	public static String TransformationEditor_tooltipHideTargetViewers;
 	public static String TransformationEditor_tooltipShowSourceVariableViewers;
 	public static String TransformationEditor_tooltipShowTargetViewers;
+	public static String TransformationEditor_unableToOpenIncompatibleTransformationFile;
```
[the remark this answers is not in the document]
lhein: sonar issue to fix

lhein:
What about the PR @apupier created on your fork? https://github.com/jpav/jbosstools-fuse/pull/3

jpav:
There were some minor changes I made to what he submitted, so I just incorporated those and the test-related code he offered into this latest PR update.

transformation/tests/org.jboss.tools.fuse.transformation.editor.tests/src/test/java/org/jboss/tools/fuse/transformation/editor/internal/util/DozerMigratorTest.java
```diff
@@ -0,0 +1,62 @@
...
+		Path expectedMigratedFile = createTmpFileFromBundle("/dozerPre6.1-postMigration.xml", "testDozerMigration-expectedResult");
+		File migratedFile = File.createTempFile("testDozerMigration", ".xml");
+
+		assertThat(new DozerMigrator().migrate(migratedFile , initialFile)).isTrue();
+
+		assertThat(migratedFile).hasSameContentAs(expectedMigratedFile.toFile());
+	}
+
+	@Test
+	public void testMigrateIfNecessaryForXsdWrittentOnSeveralLines() throws Exception {
```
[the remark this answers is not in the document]
jpav: I linked issue 2763 to test issue https://issues.jboss.org/browse/FUSETOOLS-2831

apupier:
I didn't took care of that but the description of the JIRA issue/commit message and the content are not matching.
The description/commit message is mentioning "Provide automatic migration of dozer file when switching camel version from < 2.20 to >= 2.20" although what is implemented in this PR  and commits is "Provide automatic migration of Dozer file when opening Data Transformation editor" which is quite different. Especially for people, just switching Camel version without openign a Data Transoformation editor, it will fail at Runtime

PR #1268, opened 2018-02-05
FUSETOOLS-2769 - Cache ICamelService

apupier:
- it is saving 20 ms on each call
- loosing dynamicity in case a service is added/removed during runtime
- but I think that it is no more used anyway and the camel.version used
is always "dynamic" and I'm pretty sure that nobody is using this
dynamicity

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?
- [ ] Are there integration tests (or at least a jira to tackle these)? mainly removing code and there are already several tests going through the code changed

## Legal


apupier:
so for instance when adding a File component to a diagram, the gain is of 120 ms

before:
image

and after I can't even show it because the time is too low to be captured in diagram with sampler mode used

apupier:
@lhein should we stay with the "cache" way? or completely remove the Service stuff which I think is not used anymore (since Camel Version decoupling)?

lhein:
That is a relict from when we had separate bundles for each camel version. We can remove/refactor that imo.

PR #1281, opened 2018-02-20
Replace the synchronized class "StringBuffer" by an unsynchronized one

apupier:
such as "StringBuilder".

and fix some other Sonar reported issues on same classes

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?

## Legal


apodhrad:
Hi @apupier, I'm just wondering about the performance. Is the difference significant? I have never checked it.
Nevertheless, it is a good point - the synchronization is not needed here.

apupier:
> I'm just wondering about the performance. Is the difference significant?

In our case, i think not that much but not tested.
I prefer to not take risk to have performance issue one day because of that and it helps to have cleaner Sonar issues, so we can focus on more important ones after (and more easily detect new ones)

PR #1337, opened 2018-04-06
FUSETOOLS-2874 - embed some Camel Catalog

apupier:
- embed catalog for latest versions to avoid time of downloading
artifacts for Groovy/Ivy which can be very slow due to several bugs
still not fixed upstream
- latest versions are: 2.20.3, 2.21.0 (reusing the camel.core.impl plugin), latest 6.3.0, latest 6.2.1 and latest FIS 2.0
- there is some duplicated code but I think we are a bit forced to avoid
classpath issue as we provide the same class in different version.

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [x] Are there unit-tests?
- [x] Are there integration tests (or at least a jira to tackle these)?

## Legal


core/plugins/org.fusesource.ide.camel.model.service.impl/src/org/fusesource/ide/camel/model/service/internal/DynamicCamelCatalog.java
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	@Override
+	public String getLoadedVersion() {
+		return catalog.getLoadedVersion();
+	}
+
+	public String loadVersion(String requestedVersion) {
+		if (!catalog.loadVersion(requestedVersion)) {
+			if(OLDER_CAMEL_VERSION_WITH_CATALOG.compareTo(requestedVersion) > 1) {
```
lhein: pretty sure we had a method somewhere that does this check already

apupier: yes, it was in CamelService :-)
I moved this code from CamelService class: https://github.com/jbosstools/jbosstools-fuse/pull/1337/files#diff-6f3995e9cde8ad715b8ba14e9d12872aL113

or there was duplicated before that we can factorize?

lhein: I think we have some duplicated code around camel version comparisons. How about moving Camel version specific util code to https://github.com/jbosstools/jbosstools-fuse/blob/master/core/plugins/org.fusesource.ide.foundation.core/src/org/fusesource/ide/foundation/core/util/VersionUtil.java instead?

apupier: updated this old code to use the VersionUtil

lhein:
Aurelien, can u raise a Jira to rename versionutil to CamelVersionUtils and to clean duplicated code for version comparison in other classes?

apupier:
> Aurelien, can u raise a Jira to rename versionutil to CamelVersionUtils and to clean duplicated code for version comparison in other classes?

VersionUtil can be used for all Maven versions , so VersionUtil name seems better.
I opened a task for reusing the Util class: https://issues.jboss.org/browse/FUSETOOLS-2939 even if I do'n think there is a great value to go through hall code for that. When changing related code, to update to use it seems a better strategy for me.

PR #1348, opened 2018-04-16
FUSETOOLS-2879 - filter runtime in "New Runtime" wizard

apupier:

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?
- [x] Are there integration tests (or at least a jira to tackle these)?

## Legal

### DIFF
--- a/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/FuseIntegrationProjectWizardRuntimeAndCamelPage.java
+++ b/editor/plugins/org.fusesource.ide.projecttemplates/src/org/fusesource/ide/projecttemplates/wizards/pages/FuseIntegrationProjectWizardRuntimeAndCamelPage.java
@@ -19,6 +19,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.eclipse.core.databinding.DataBindingContext;
 import org.eclipse.core.databinding.beans.PojoProperties;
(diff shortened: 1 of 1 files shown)


apupier:
for Karaf:
image
for wildly/EAP:
image
for comparison without filter (as it is currently):
image
image


we can notice that several runtimes not Fuse/Karaf/Wildfly/EAP are still appearing but it seems a bug to me. What do you think? @robstryker

apupier:
@lhein @bfitzpat having the list hardcoded makes it easier to see which versions we are supporting and is a bit faster, on the other hand when a new runtime is added, we will need to update this list. Do you have another argument to choose between one those 2 solutions? (we can also start like that and modify in a next iteration)

lhein:
doesn't it work with org.jboss.ide.eclipse.as.runtime.* also? I would prefer if we can make it more dynamic

apupier:
> doesn't it work with org.jboss.ide.eclipse.as.runtime.* also? I would prefer if we can make it more dynamic

no it doesn't work

lhein:
Then I fear I don't understand what the alternative approach to hard coding would be here?

apupier:
> Then I fear I don't understand what the alternative approach to hard coding would be here?

Retrieve all runtimes from our code and filter the list based on ids and the kind of pattern that you mentioned.

lhein:
Actually the later approach is easier to maintain and I would opt for the dynamic approach.

robstryker:
The weird ones still being shown aren't server adapters. They are references to extensions that can be installed. I imagine an API enhancement could be made here to not show those.

apupier:
> The weird ones still being shown aren't server adapters. They are references to extensions that can be installed. I imagine an API enhancement could be made here to not show those.

thanks for the answer. I opened the enhancement request: https://issues.jboss.org/browse/JBIDE-25920

apupier:
> Actually the later approach is easier to maintain and I would opt for the dynamic approach.

updated with dynamic approach.
Maybe some old Wildfly/EAP versions can be filtered out? If yes, we can raise another JIRA enhancement request.

PR #1388, opened 2018-05-29
FUSETOOLS-2999 - fixing validation

bfitzpat:

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?

## Legal

### DIFF
--- a/editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/pages/Wsdl2RestWizardBasePage.java
+++ b/editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/pages/Wsdl2RestWizardBasePage.java
@@ -160,6 +160,25 @@ public abstract class Wsdl2RestWizardBasePage extends WizardPage {
 				new UpdateValueStrategy().setBeforeSetValidator(validator),
 				null);
 	}
+	
+	/**
+	 * Creates a databinding Binding between a Text control, a model ID, and validator that employs a
+	 * 500ms delay when typing so we wait until there's a pause in typing.
+	 * @param control
+	 * @param modelID
+	 * @param validator
+	 * @return
+	 */
+	protected Binding createBindingWithTypeDelay(Text control, String modelID, IValidator validator) {
+		IObservableValue<?> wsdlTarget = WidgetProperties.text(SWT.Modify).observeDelayed(500, control);
+		@SuppressWarnings("unchecked")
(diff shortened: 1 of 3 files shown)


bfitzpat:
Will try and repro here. Thanks.

bfitzpat:
Can definitely reproduce here. I'm adding a type delay (WidgetProperties.text(SWT.Modify).observeDelayed(500, control)) to help with it a bit, but the ultimate issue seems to come from a SAX parsing issue with the WSDL referenced at the URL you're trying...
```
WSDLException: faultCode=PARSER_ERROR: Problem parsing 'http://nullege.com/codes/search/SOAPpy.WSDL'.: org.xml.sax.SAXParseException: The reference to entity "f" must end with the ';' delimiter.
WSDLException: faultCode=PARSER_ERROR: Problem parsing 'http://nullege.com/codes/search/SOAPpy.WSDL'.: org.xml.sax.SAXParseException: The reference to entity "f" must end with the ';' delimiter.
```

Any suggestions on how to better/more efficiently process the contents of the URL? Ultimately there are two parts to this validation - one simply ensures that it's not empty and is a valid URL; two then tries to parse the WSDL to find the service location. It's this second part where we're getting hung up in WSDL2RestWizardBasePage.getLocationFromWSDL().

@apupier

apupier:
> Any suggestions on how to better/more efficiently process the contents of the URL?

My first suggestion would be to validate only when it is empty as a first iteration.
To validate if the wsdl is valid, I would say that the validation needs to be launched asynchronously (but makes message status harder to handle)
In "New Fuse Integration Wizard", the validation of camel version can take time, it is then done on-demand via button and when clicking finish.

bfitzpat:
Hmm. I'll take a look at it again today. Thanks for the suggestions.

bfitzpat:
With the validation tweak in place, parsing becomes VERY fast and the issue goes away.

apupier:
Now the validation is just checking hat the URL is valid

image

can be enough for a first iteration

bfitzpat:
Even before, we didn't fully validate the WSDL, only that the field was not blank and the URL was accessible. Are you proposing a third validation where we somehow validate that it's valid WSDL or a valid XML file? That would be a great check, but I'm honestly not sure that the time required would be worth the cost. The SAX parser (without the validation tweaks I made yesterday) takes several seconds to throw back an invalid result for some reason and would still show up as potentially "hung" if we went that route.

apupier:
> Are you proposing a third validation where we somehow validate that it's valid WSDL or a valid XML file?

that's a possibility. Maybe not worthy enough given time to develop it.

I also wanted to check that I understood the expected behavior correctly.

PR #1397, opened 2018-06-04
JBDS-4706 rename JBDS to Red Hat Developer...

nickboldt:
JBDS-4706 rename JBDS to Red Hat Developer Studio


apupier:
@nickboldt if I understood well, the test will pass when all Jboss Tools project would have been updated and available when building Fuse Tooling, right?
Do you know which project we should monitor to relaunch the build?

nickboldt:
Why do these string changes depend on JBT projects being updated and built?

apupier:
> Why do these string changes depend on JBT projects being updated and built?

the test is going through a dialog having the JBT name in the title. I guess there is no specific jboss tools red deer dialog for it. @tsedmik Can you check please? is it worthy contribute it if not?

tsedmik:
@apupier yes, as far as I know, there is no such dialog implementation in RedDeer.

PR #1416, opened 2018-06-29
FUSETOOLS-2945 - automatically assign ID values to rest elements

lhein:

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?

## Legal

### DIFF
--- a/editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
+++ b/editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
@@ -11,8 +11,10 @@
 package org.fusesource.ide.wsdl2rest.ui.wizard;
 
 import java.io.File;
+import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.file.Path;
 import java.util.ArrayList;
@@ -33,22 +35,27 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
+import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jface.dialogs.ErrorDialog;
 import org.eclipse.jface.dialogs.IMessageProvider;
-import org.eclipse.jface.operation.IRunnableWithProgress;
 import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.jface.wizard.Wizard;
+import org.eclipse.swt.widgets.Display;
 import org.eclipse.ui.INewWizard;
 import org.eclipse.ui.IPageLayout;
 import org.eclipse.ui.ISelectionService;
 import org.eclipse.ui.IWorkbench;
+import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PlatformUI;
+import org.fusesource.ide.camel.editor.CamelEditor;
+import org.fusesource.ide.camel.editor.utils.CamelUtils;
 import org.fusesource.ide.camel.editor.utils.MavenUtils;
 import org.fusesource.ide.camel.model.service.core.io.CamelIOHandler;
 import org.fusesource.ide.camel.model.service.core.util.CamelCatalogUtils;
 import org.fusesource.ide.camel.model.service.core.util.CamelMavenUtils;
 import org.fusesource.ide.foundation.core.util.Strings;
+import org.fusesource.ide.projecttemplates.util.BasicProjectCreatorRunnableUtils;
 import org.fusesource.ide.wsdl2rest.ui.internal.UIMessages;
 import org.fusesource.ide.wsdl2rest.ui.internal.Wsdl2RestUIActivator;
 import org.fusesource.ide.wsdl2rest.ui.wizard.pages.Wsdl2RestWizardFirstPage;
@@ -72,6 +79,7 @@ public class Wsdl2RestWizard extends Wizard implements INewWizard {
 
 	private Wsdl2RestWizardSecondPage pageTwo;
 	
+	private boolean done = false; 
 	private boolean inTest = false;
 	
 	/**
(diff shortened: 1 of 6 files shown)


lhein:
I also modified the code to open up the newly created camel route in the editor. Not having it open on perform finish on the wizard felt somewhat wrong.

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
```diff
@@ -283,8 +291,8 @@ private IResource getFileForPath(IProject project, IPath path, IProgressMonitor
 	 * @param monitor
 	 * @throws Exception
 	 */
-	private void generate(IProgressMonitor monitor) throws Exception {
-		SubMonitor subMon = SubMonitor.convert(monitor, 5);
+	private void generate(IProgressMonitor monitor) throws IOException, CoreException, URISyntaxException {
+		SubMonitor subMon = SubMonitor.convert(monitor, 6);
```
apupier: I think that it should stay to 5, can you double-check?

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
```diff
@@ -299,37 +307,67 @@ private void generate(IProgressMonitor monitor) throws Exception {
...
+			Thread.currentThread().setContextClassLoader(oldLoader);
+			project.refreshLocal(IResource.DEPTH_INFINITE, subMon.split(1));
+		}
+		if (contextPath != null) {
+			Job job = new Job("Update Camel Context File...") {
+				
+				@Override
+				protected IStatus run(IProgressMonitor monitor) {
+					SubMonitor subMon = SubMonitor.convert(monitor, 1);
+					while (!done) {
```
apupier: I fear that if an exception is throw the done which is not in a finally block will stay at false and this job will loop forever

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
```diff
@@ -299,37 +307,67 @@ private void generate(IProgressMonitor monitor) throws Exception {
...
+						try {
+							Thread.sleep(1000);
+						} catch (InterruptedException e) {
+							Thread.currentThread().interrupt();
+						}
+					}
+					Display.getDefault().asyncExec( () -> updateIDValueInCamelFile(contextPath, subMon.split(1)));
```
apupier: modifying values asynchronuously seems dangerous, what happens if the operation is taking a long tie eand the user has already started to modify it?

lhein:
If you look at my changes you will find out that both changes are very tight coupled.

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
bfitzpat: Kind of agree here, since this is already wrapped in a getContainer().run(...) call. I would worry about async vs. sync.

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
```diff
@@ -370,6 +408,59 @@ protected Path initContextForTool(boolean isBlueprint, boolean isSpringBoot, Fil
 		return contextpath;
 	}
 
+	/**
+	 * 
+	 * @param path
+	 * @param monitor
+	 */
+	protected void updateIDValueInCamelFile(Path path, IProgressMonitor monitor) {
```
bfitzpat: I like this method - a bit brute force, but simply opening the file and saving it is all we need.

bfitzpat:
Tests fine locally though, so I like the change in principle. :)

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
lhein: If you look at line 310 you will find a .split(2). So 6 is correct.

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
lhein: agreed

lhein:
> there are few parts which seems not very robust in case of concurrency problems or unexpected runtime exception
    in case there is an existing camel route opened in the Camel route editor with Source tab (which will be overwritten), a popup asking if we want to overwrite what is on filesystem is shown (which is fine) but in this case, the ids are not applied. They are applied the next time the editor is loaded.
    in case there is an existing camel route opened in the Camel route editor with Design/Rest tab (which should be overwritten) --> the overwrite is not working, still the old file

I fixed the "done" issue by putting it into the finally block calling the generate method. For the other 2 issues I decided to close the editor if its currently open and reopen it afterwards.  On a modern system the user won't even recognize it except that the tab switches to design tab. But given the complexity of the problems that is the most clean solution I could think of. The changes are discarded when closing the editor because the whole file will be regenerated anyways.

bfitzpat:
Once we have a green build and are ready to start work on 11.1.0.AM1, this looks good to go.

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
apupier: I don't like this method which is making editors blinking and changing the position of them in case several are open. It is also the root cause of not working when an xml editor is opened for the camel file.

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
```diff
@@ -299,37 +310,66 @@ private void generate(IProgressMonitor monitor) throws Exception {
...
+			Job job = new Job("Update Camel Context File...") {
+				
+				@Override
+				protected IStatus run(IProgressMonitor monitor) {
+					SubMonitor subMon = SubMonitor.convert(monitor, 1);
+					while (!done) {
+						try {
+							Thread.sleep(1000);
+						} catch (InterruptedException e) {
+							Thread.currentThread().interrupt();
```
apupier: shouldn't we stop the loop in case of Interruption?

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
lhein: I tried to address your concerns. Please let me know if that fits better now.

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
lhein: added a stop logic

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
```diff
@@ -370,6 +411,64 @@ protected Path initContextForTool(boolean isBlueprint, boolean isSpringBoot, Fil
...
+				ed.getDesignEditor().getModel().registerDOMListener();
+				ed.doSave(subMon.split(1));
+				if (ed.getActivePage() == CamelEditor.REST_CONF_INDEX) {
+					ed.getRestEditor().reload();
+				} else if (ed.getActivePage() == CamelEditor.SOURCE_PAGE_INDEX) {
+					ed.switchToSourceEditor();
+				}
+			} else {
+				io.saveCamelModel(cf, f, subMon.split(1));
+				openCamelContextFile(cf.getResource(), monitor);
```
apupier: need to use the submonitor here too

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
apupier: the model is not modified now in design view (i can see it is modified in the Project explorer) and when I switch to the Source tab, the previous model is overriding the generated one

apupier:
it has been implemented in wsdl2rest utility side, closing

PR #1478, opened 2018-09-20
FUSETOOLS-2869 - adding more dependencies via wsdl2rest wizard

bfitzpat:

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?

## Legal

### DIFF
--- a/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/component/wizard/ComponentManager.java
+++ b/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/component/wizard/ComponentManager.java
@@ -13,6 +13,7 @@ package org.fusesource.ide.camel.editor.component.wizard;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -73,6 +74,12 @@ public class ComponentManager {
 
 	public Set<Component> getAllComponents() {
 		return new HashSet<>(model.getComponents());
+	}
+	
+	public Component getComponentById(String id) {
+		return model.getComponents().stream()
+				.filter(component -> id.equalsIgnoreCase(component.getId())).findAny()
+				.orElse(null);		
 	}
 
 }
--- a/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/restconfiguration/RestConfigEditor.java
+++ b/editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/restconfiguration/RestConfigEditor.java
@@ -10,6 +10,7 @@
  ******************************************************************************/
 package org.fusesource.ide.camel.editor.restconfiguration;
 
+import java.lang.reflect.InvocationTargetException;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.Iterator;
@@ -54,11 +55,14 @@ import org.eclipse.swt.widgets.Text;
 import org.eclipse.swt.widgets.ToolBar;
 import org.eclipse.ui.IEditorInput;
 import org.eclipse.ui.IEditorSite;
+import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.PartInitException;
+import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.forms.widgets.FormToolkit;
 import org.eclipse.ui.forms.widgets.ScrolledForm;
 import org.eclipse.ui.forms.widgets.Section;
 import org.eclipse.ui.part.EditorPart;
+import org.eclipse.ui.progress.IProgressService;
 import org.fusesource.ide.camel.editor.CamelEditor;
 import org.fusesource.ide.camel.editor.internal.CamelEditorUIActivator;
 import org.fusesource.ide.camel.editor.internal.UIMessages;
@@ -92,7 +96,7 @@ public class RestConfigEditor extends EditorPart implements ICamelModelListener,
 
 	private static final String OFF = "off"; //$NON-NLS-1$
 	
-	private CamelEditor parentEditor;
+	CamelEditor parentEditor;
 	private Composite parent;
 	private ScrolledForm form;
 	private FormToolkit toolkit;
@@ -490,6 +494,7 @@ public class RestConfigEditor extends EditorPart implements ICamelModelListener,
 			public void widgetSelected(SelectionEvent e) {
 				if (rce != null) {
 					rce.setComponent(componentCombo.getText());
+					addComponentDependency(componentCombo.getText());
 				}
 			}
 		});
@@ -547,6 +552,19 @@ public class RestConfigEditor extends EditorPart implements ICamelModelListener,
 		section.setClient(client);
 		
 		return client;
+	}
+	
+	private void addComponentDependency(String componentName) {
+		IWorkbench wb = PlatformUI.getWorkbench();
+		IProgressService ps = wb.getProgressService();
+		try {
+			ps.busyCursorWhile(new AddComponentDependencyRunnable(this, componentName));
+		} catch (InvocationTargetException e) {
+			CamelEditorUIActivator.pluginLog().logError(e);
+		} catch (InterruptedException e) {
+			CamelEditorUIActivator.pluginLog().logError(e);
+			Thread.currentThread().interrupt();
+		}
 	}
 
 	private void clearUI() {
--- a/editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
+++ b/editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
@@ -61,6 +61,8 @@ import org.jboss.fuse.wsdl2rest.impl.Wsdl2Rest;
  */
 public class Wsdl2RestWizard extends Wizard implements INewWizard {
 
+	private static final String APACHE_CAMEL_GROUP_ID = "org.apache.camel"; //$NON-NLS-1$
+
 	/**
 	 * Collection of settings used by the wsdl2rest utility.
 	 */
@@ -172,19 +174,32 @@ public class Wsdl2RestWizard extends Wizard implements INewWizard {
 	 * @return
 	 */
 	public boolean isProjectBlueprint(IProject project) {
+		return projectHasDependency(project, APACHE_CAMEL_GROUP_ID, "camel-blueprint"); //$NON-NLS-1$ //$NON-NLS-2$
+	}
+	
+	/**
+	 * Checks for blueprint dependency
+	 * @param project
+	 * @return
+	 */
+	public boolean isProjectSpring(IProject project) {
+		return projectHasDependency(project, APACHE_CAMEL_GROUP_ID, "camel-spring"); //$NON-NLS-1$ //$NON-NLS-2$
+	}
+
+	private boolean projectHasDependency(IProject project, String groupId, String artifactId) {
 		CamelMavenUtils cmu = new CamelMavenUtils();
 		List<Dependency> projectDependencies = cmu.getDependencyList(project);
 		Iterator<Dependency> depIter = projectDependencies.iterator();
 		while(depIter.hasNext()) {
 			Dependency dependency = depIter.next();
-			if ("org.apache.camel".equals(dependency.getGroupId()) && //$NON-NLS-1$
-					"camel-blueprint".equals(dependency.getArtifactId())) { //$NON-NLS-1$
+			if (groupId.equals(dependency.getGroupId()) &&
+					artifactId.equals(dependency.getArtifactId())) { 
 				return true;
 			}
 		}
 		return false;
 	}
-	
+
 	/**
 	 * Checks for spring boot
 	 * @param project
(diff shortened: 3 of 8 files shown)


bfitzpat:
This is just the beginning of this effort. Changes still need to be made to the Rest tab to support adding other dependencies when the rest configuration component changes.

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
```diff
@@ -196,18 +196,66 @@ public boolean isProjectSpringBoot(IProject project) {
...
+		org.fusesource.ide.camel.model.service.core.catalog.Dependency jaxrsSpec = 
+				makeDependency("org.jboss.spec.javax.ws.rs",  //$NON-NLS-1$
+						"jboss-jaxrs-api_2.0_spec",  //$NON-NLS-1$
+						"1.0.0.Final-redhat-1"); //$NON-NLS-1$
+		deps.add(jaxrsSpec);
+		
+		org.fusesource.ide.camel.model.service.core.catalog.Dependency camelJackson = 
+				makeDependency("org.apache.camel",  //$NON-NLS-1$
+						"camel-jackson",  //$NON-NLS-1$
+						"${camel.version}"); //$NON-NLS-1$
```
apupier: I think we can't rely on the fact that camel.version property is defined.

bfitzpat: Then what do you recommend? We also can't rely on the fact that the BOM is defined. Is there a way to get the version of the core camel dependency from the maven utilities?

apupier: several cases are handled in org.fusesource.ide.projecttemplates.util.maven.MavenUtils.updateCamelVersionDependencies(Model, List<Dependency>, String) for instance.

To retrieve Camel version there is org.fusesource.ide.camel.model.service.core.util.CamelMavenUtils.getCamelVersionFromMaven(IProject, boolean)

bfitzpat: Perfect. Thanks

bfitzpat:
@lhein and @apupier I have updated this PR to include functionality to add the Rest Configuration component dependency when the user changes the component on the Rest tab of the editor. But I'm not happy with this approach for two reasons: 

1. It takes a while and doesn't pop up a progress dialog (NullProgressMonitor is used). Is there a better way to handle this so that we popup a dialog or handle this during the save? I feel like this is similar to how we update the project when the user drops a new component on the graphical editor and update the pom, but want to make sure there isn't a better way to handle it.
2. Are there hidden dependencies that go along with this? If so, how do we use the camel catalog to gather and add them all at once? I was unable to quickly find an example of how to use the catalog to do this, but suspect it's a better way to go in the long run. 

This is all well ahead of the next release, but I want to try and get these ducks in a row so they are ready to go when we can get them in and tested early in the process of the next release.

Thanks.

apupier:
> It takes a while and doesn't pop up a progress dialog (NullProgressMonitor is used). Is there a better way to handle this so that we popup a dialog or handle this during the save? I feel like this is similar to how we update the project when the user drops a new component on the graphical editor and update the pom, but want to make sure there isn't a better way to handle it.

Several possibilities to have a spinning cursor/progress monitor:
- https://wiki.eclipse.org/FAQ_Why_should_I_use_the_new_progress_service%3F
- https://wiki.eclipse.org/FAQ_How_do_I_switch_from_using_a_Progress_dialog_to_the_Progress_view%3F
- https://wiki.eclipse.org/E4/EAS/Progress_Service

I'm not sure how you imagine to handle it on save.
When creating the rest dsl , having the pom not modified but on each save check for dependencies and see if some needs to be added? If yes, it seems to not be very convenient as the user will need to wait everytime he/she is calling save. Also if we have a bug, the library will be added on every save which will be a blocker although if it is during the first generation , it can be removed and there is an easy workaround.
Or you have another way in mind?

> Are there hidden dependencies that go along with this? If so, how do we use the camel catalog to gather and add them all at once? I was unable to quickly find an example of how to use the catalog to do this, but suspect it's a better way to go in the long run.

Hidden from what?
What we discussed is that the dependencies are in the Camel catalog.
For instance with dataformat, there is a getDependencies() method available, see  usage org.fusesource.ide.camel.editor.globalconfiguration.dataformat.wizards.NewDataFormatWizard.createDataFormatNode(DataFormat, String, IProgressMonitor)
it is also used here for Camel components: org.fusesource.ide.camel.editor.features.create.ext.AbstractComponentBasedCreateFigurefeature.createNode(AbstractCamelModelElement, boolean) (for all the palette components)

apupier:
currently compile error as it requires https://github.com/jboss-fuse/wsdl2rest/pull/62

```
18:39:27 [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:1.2.0:compile (default-compile) on project org.fusesource.ide.wsdl2rest.ui: Compilation failure: Compilation failure: 
18:39:27 [ERROR] /home/jenkins/workspace/jbosstools-fuse_PullRequest/editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java:[408] 
18:39:27 [ERROR] 	tool.setNoVelocityLog(true);
18:39:27 [ERROR] 	     ^^^^^^^^^^^^^^^^
18:39:27 [ERROR] The method setNoVelocityLog(boolean) is undefined for the type Wsdl2Rest
18:39:27 [ERROR] 1 problem (1 error)
```

bfitzpat:
Yup! There are some issues to work though. The Velocity log change has been merged upstream this morning, so we should start to see it in builds, which will get us past that speed bump. I'm working through your various suggestions/questions on the progress and dependency questions I had as well. Thank you @apupier

bfitzpat:
As soon as we have a new wsdl2rest build upstream to test against, I'll update the PR to address that issue. But I have reworked the dependency code in addComponentDependency to use the IProgressService and Catalog ComponentManager to ensure a smoother approach here. I couldn't find a simple way to search for a given component by ID (just by tag), but this seems to work ok.

@lhein and @apupier Thanks in advance for any additional feedback on this approach

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
```diff
@@ -196,18 +196,72 @@ public boolean isProjectSpringBoot(IProject project) {
...
+		deps.add(camelCxf);
+
+		if (isBlueprint) {
+			org.fusesource.ide.camel.model.service.core.catalog.Dependency camelServlet = 
+					makeDependency("org.apache.camel",  //$NON-NLS-1$
+							"camel-servlet",  //$NON-NLS-1$
+							camelVersion);
+			deps.add(camelServlet);
+		} else {
+			// change this once we get the spring default component updated in wsdl2rest
```
apupier: can you add a link the issue in wsdl2rest to track it more easily.
Maybe worth using a //TODO as there are better tooling to spot these kind of things.

editor/plugins/org.fusesource.ide.wsdl2rest.ui/src/org/fusesource/ide/wsdl2rest/ui/wizard/Wsdl2RestWizard.java
```diff
@@ -196,18 +196,72 @@ public boolean isProjectSpringBoot(IProject project) {
...
-				new org.fusesource.ide.camel.model.service.core.catalog.Dependency();
-		one.setArtifactId("jboss-jaxrs-api_2.0_spec"); //$NON-NLS-1$
-		one.setGroupId("org.jboss.spec.javax.ws.rs"); //$NON-NLS-1$
-		one.setVersion("1.0.0.Final-redhat-1"); //$NON-NLS-1$
-		deps.add(one);
 		IProject project = options.getProject();
+
+		// use project to determine if we are building a spring or blueprint project
+		boolean isBlueprint = isProjectBlueprint(project);
+		boolean isSpringBoot = isProjectSpringBoot(project);
```
apupier: I think we also have a case when it is Spring but not SpringBoot. But maybe for wsdl2rest there is nothing specific to add in this case?

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/restconfiguration/RestConfigEditor.java
```diff
@@ -548,6 +561,35 @@ public void widgetSelected(SelectionEvent e) {
 		
 		return client;
 	}
+	
+	private void addComponentDependency(String componentName) {
+		IWorkbench wb = PlatformUI.getWorkbench();
+		IProgressService ps = wb.getProgressService();
+		try {
+			ps.busyCursorWhile(new IRunnableWithProgress() {
```
apupier: I think it will be more readable if the inner class is extracted to its own class file

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/restconfiguration/RestConfigEditor.java
```diff
@@ -548,6 +561,35 @@ public void widgetSelected(SelectionEvent e) {
...
+				public void run(IProgressMonitor pm) {
+					IProject project = parentEditor.getDesignEditor().getWorkspaceProject();
+					final CamelModel camelModel = CamelCatalogCacheManager.getInstance().getCamelModelForProject(project, pm);
+					final ComponentManager componentManager = new ComponentManager(camelModel);
+					Set<Component> componentSet = componentManager.getAllComponents();
+					if (!componentSet.isEmpty()) {
```
apupier: I think finding the components by Id is a great use case that might be useful in other contexts. What do you think to create the a getComponentbyId(String id) in ComponentManager?
It will avoid to loop and have a break; statement in this RunnableWithProgress

bfitzpat:
I don't believe we have control over the "Building workspace" message, which happens any time we're updating the project (delete the camel context file or some java files and you will likely see it come up like the following dialog)

image

Not sure this is from m2e or from Eclipse itself.

bfitzpat:
Updated the PR to address latest round of feedback from @apupier

bfitzpat:
rebased on latest master changes to see what shakes out now.

bfitzpat:
Rebased after the fixes for FUSETOOLS-3163 went in -- this fixes the wsdl2rest tests locally, now will hope for the same result on Jenkins. :)

bfitzpat:
With tests now passing and (hopefully) the sonar issues resolved, I think we're in good shape for this one.

bfitzpat:
retest this please

bfitzpat:
@apupier @lhein  

Seems the only thing we have left now is the version of wsdl2rest to use. Should we ask Jon & Tom to stash the 0.8.0.fuse-720027 version for now and create a follow-up issue to update it when we have a more permanent version or another temporary one we want to use?

apupier:
> to stash the 0.8.0.fuse-720027 version for now and create a follow-up issue to update it when we have a more permanent version or another temporary one we want to use?

the thing is that the wsdl version needs to be stashed but also all the dependencies (which can have different versions) of this specific version. Last time I asked for that they didn't do it.

bfitzpat:
> 
> 
> > to stash the 0.8.0.fuse-720027 version for now and create a follow-up issue to update it when we have a more permanent version or another temporary one we want to use?
> 
> the thing is that the wsdl version needs to be stashed but also all the dependencies (which can have different versions) of this specific version. Last time I asked for that they didn't do it.

Yes - I have opened another JIRA for managing the wsdl2rest version here - https://issues.jboss.org/browse/FUSETOOLS-3164

bfitzpat:
Updated to 0.8.0.fuse-720034 for the wsdl2rest version. Will change to the Fuse productized version as denoted by https://issues.jboss.org/browse/FUSETOOLS-3164 when it becomes available. This should be the last change to get these fixes into master if all tests pass successfully.

@apupier @lhein

bfitzpat:
retest

bfitzpat:
testPR

nickboldt:
testPR

(When in danger or in doubt, 
Run in circles, scream and shout.
Then do try the build again, 
or turn it off and on again.)

bfitzpat:
Bah humbug.

editor/plugins/org.fusesource.ide.wsdl2rest/pom.xml
```diff
@@ -105,6 +108,17 @@
 			<groupId>log4j</groupId>
 			<artifactId>log4j</artifactId>
 		</dependency>
+        <dependency>
```
apupier: indentation issue

bfitzpat: The indention is not causing the build failures. :)

apupier: yes but it needs to be fixed before merging

editor/plugins/org.fusesource.ide.camel.editor/src/org/fusesource/ide/camel/editor/component/wizard/ComponentManager.java
```diff
@@ -74,5 +75,16 @@ private void initTagMap(CamelModel model) {
 	public Set<Component> getAllComponents() {
 		return new HashSet<>(model.getComponents());
 	}
+	
+	public Component getComponentById(String id) {
+		Iterator<Component> iter = model.getComponents().iterator();
```
apupier: optional: it i s possible to write it with lambda  with fewer lines of code
something like 
```
return model.getComponents().stream()
				.filter(component -> id.equalsIgnoreCase(component.getId())).findAny()
				.orElse(null);
```

bfitzpat:
@apupier Updated per your suggestions

PR #1508, opened 2019-02-28
FUSETOOLS-3185 - Avoid 0.9.0 which is introducing regression

apupier:
when a new version without regression will be available, we will just
need to move to this new version

https://issues.jboss.org/browse/JBIDE-26563 needs to be fixed before this PR can build (and so be merged)

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?

## Legal

### DIFF
--- a/editor/plugins/org.fusesource.ide.project/META-INF/MANIFEST.MF
+++ b/editor/plugins/org.fusesource.ide.project/META-INF/MANIFEST.MF
@@ -38,7 +38,7 @@ Require-Bundle: org.eclipse.ui;bundle-version="3.106.0",
  org.fusesource.ide.camel.validation;bundle-version="10.0.0",
  org.fusesource.ide.camel.editor;bundle-version="10.0.0",
  org.eclipse.jst.server.core;bundle-version="1.2.400",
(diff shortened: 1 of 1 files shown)


apupier:
> This will disable the plugin only, right?

it means that we must have a org.sonatype.tycho.m2e with a version between 0.8.1 and 0.9.0.
In case, both versions are provided, the behavior needs to be checked.

lhein:
I just meant that when we only have 0.9.x in the TP our plugin will not resolve any more and be disabled.

apupier:
> In case, both versions are provided, the behavior needs to be checked.

if both provided, the MANIFEST.MF is generated so it is fine.

apupier:
> I just meant that when we only have 0.9.x in the TP our plugin will not resolve any more and be disabled.

no, the build is breaking. And you won't be able to install it

lhein:
ok, that is at least a stable behaviour.

apupier:
blocked by https://github.com/jbosstools/jbosstools-build/pull/287

PR #1515, opened 2019-03-15
Revert "FUSETOOLS-3187 - ignore failing test"

apupier:
Reverts jbosstools/jbosstools-fuse#1510

new Target Platform with RedDeer 2.5.0 containing the fix available https://dev-platform-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Devstudio/view/devstudio_master/job/jbosstoolstargetplatform_master/54/


djelinek:
there is also have to be updated reddeer version in UI tests plugins

apupier:
> there is also have to be updated reddeer version in UI tests plugins

right, it will be cleaner. I will update the PR

PR #1516, opened 2019-03-18
FUSETOOLS-3190 - use CamelFile method to have xml to save

apupier:
with java 11, the methods previously used with transformers is
introducing extra-lines on each lines which is breaking the field
containing the CDATA.

note that the saveAsCamelModel seems to be used only in test for real.
Other places referenced:
- Datatransformation, it is used only when the class field saveModel is
set to true which is never the case currently.
- CamelEditorPersistencyDiagram: currently not used, the saved is
intercepted at multieditor level, synchronizing the model with the
source tab and then saving from source tab.

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?

## Legal


apupier:
> Do we lose the nice formatting/indent capabilities by going with this new approach?

it is the opposite, we are keeping it :-)
it is now using the Eclipse preferences for the formatting choice

PR #1540, opened 2019-04-19
FUSETOOLS-3186 - Use cached JSON ObjectReader

apupier:

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?

## Legal


apupier:
before
image

after
image

with 3 different camel version used in the workspace, we are going from 2300ms to 400ms (5.75 times faster!)

bfitzpat:
The time savings is impressive!

apupier:
> Obviously there's an issue with a test build somewhere?

I think it is due to flaky test. I relaunched

> But is there ever a case where we need to reset the language reader? Or are we at most only loading the catalog a single time during a session with no way to re-load the catalog without shutting down the IDE and restarting it?

The Object Reader which is cached is for a specific metamodel. We are providing a single one for each of the Component, DataFormat, Language, EIP.
The same metamodel is used to load any versions of the Camel Catalog. No need to restart the IDE.
The day, the metamodel used to load the Camel catalog won't be backward compatible, we will have to generate different metamodel classes (Component, DataFormat, Language, EIP) but as we put the ObjectReader inside the classes, a new ObjectReader will be available quite naturally.

apupier:
For information (and if you can notice others spot for improvements), reusing the same ObjectReader for the metamodel class is recommended in the wiki of the library: https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance#basics-things-you-should-do-anyway

PR #1564, opened 2019-07-10
FUSETOOLS-3245 - ensuring that we always insert a new rest

bfitzpat:
element it goes after the Rest Configuration node.

- tested in the case with no route, elements added correctly
- tested in the case with an existing route, elements added correctly

# Pull Request Checklist

## General

## Functional

## Maintainability

## Tests

- [ ] Are there unit-tests?

## Legal

### DIFF
--- a/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/CamelContextElement.java
+++ b/core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/CamelContextElement.java
@@ -423,8 +423,9 @@ public class CamelContextElement extends CamelRouteContainerElement {
 			Node currentNode = childNodes.item(i);
 			if (currentNode.getNodeType() == Node.ELEMENT_NODE) {
 				Element currentElement = (Element) currentNode;
-				if (!DATA_FORMATS_NODE_NAME.equals(currentElement.getTagName()) &&
-						!ENDPOINT_NODE_NAME.equals(currentElement.getTagName())) {
+				String tagNameWithoutPrefix = CamelUtils.getTagNameWithoutPrefix(currentElement);
+				if (!DATA_FORMATS_NODE_NAME.equals(tagNameWithoutPrefix) &&
+						!ENDPOINT_NODE_NAME.equals(tagNameWithoutPrefix)) {
 					return currentNode;
 				}
 			}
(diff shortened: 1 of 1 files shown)


bfitzpat:
Before this change, it only worked in the case where an element existed in the context. Now it works in that case as well as the case where no route exists.

core/plugins/org.fusesource.ide.camel.model.service.core/src/org/fusesource/ide/camel/model/service/core/model/CamelContextElement.java
```diff
@@ -518,7 +518,16 @@ public void addRestElement(AbstractCamelModelElement def) {
 		} else {
 			// if it's already there, it's probably in the wrong place - have to move it
 			getXmlNode().removeChild(def.getXmlNode());
-			getXmlNode().insertBefore(def.getXmlNode(), getRestElementEntryPoint());
+			Node entryPoint = getRestElementEntryPoint();
+			if (entryPoint != null) {
+				Element entryPointElement = (Element) entryPoint;
+				if (REST_CONFIGURATION_NODE_NAME.equals(entryPointElement.getTagName())) {
```
apupier: it needs to be compared to (I think) the localName, not the tagname.
it is working with the tagname only if there is no namespace prefix available.
I tried to change to getLocalName but it is working only if we saved the file, it seems the node is not correctly initialized when w have just created it.

the use case, start with
```
<?xml version="1.0" encoding="UTF-8"?>
<blueprint xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="              http://www.osgi.org/xmlns/blueprint/v1.0.0 https://www.osgi.org/xmlns/blueprint/v1.0.0/blueprint.xsd              http://camel.apache.org/schema/blueprint http://camel.apache.org/schema/blueprint/camel-blueprint.xsd">
    <bean class="org.apache.aries.simple.Account" id="accountOne"/>
    <bean class="org.apache.aries.simple.Account2" id="accountTwo">
        <argument value="1"/>
    </bean>
    <bean class="org.apache.aries.simple.Account3" id="accountThree">
        <argument value="1"/>
        <property name="description" value="#1 account"/>
    </bean>
    <c:camelContext id="cbr-example-context" xmlns:c="http://camel.apache.org/schema/blueprint">
        <c:restConfiguration bindingMode="off" host="localhost"/>
    </c:camelContext>
</blueprint>
```

then go to rest tab and click to add a rest element, it is added at the beginning (and without the namespace prefix but this is another bug that I reported here https://issues.jboss.org/browse/FUSETOOLS-3264 )
```
<?xml version="1.0" encoding="UTF-8"?>
<blueprint xmlns="http://www.osgi.org/xmlns/blueprint/v1.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="              http://www.osgi.org/xmlns/blueprint/v1.0.0 https://www.osgi.org/xmlns/blueprint/v1.0.0/blueprint.xsd              http://camel.apache.org/schema/blueprint http://camel.apache.org/schema/blueprint/camel-blueprint.xsd">
    <bean class="org.apache.aries.simple.Account" id="accountOne"/>
    <bean class="org.apache.aries.simple.Account2" id="accountTwo">
        <argument value="1"/>
    </bean>
    <bean class="org.apache.aries.simple.Account3" id="accountThree">
        <argument value="1"/>
        <property name="description" value="#1 account"/>
    </bean>
    <c:camelContext id="cbr-example-context" xmlns:c="http://camel.apache.org/schema/blueprint">
        <rest id="rest-b0b6c8b0-a0a7-4705-be8b-ef4a08bb3ae7"/>
        <c:restConfiguration bindingMode="off" host="localhost"/>
    </c:camelContext>
</blueprint>
```

apupier: or use something similar to the getTagNameWithoutPrefix() which seems to workaround the bad initialization

bfitzpat: Good suggestion. Made the update, tested locally and all seems well.

apupier: tested also and it is working well (with still https://issues.jboss.org/browse/FUSETOOLS-3264 but as said before it doesn't come from this PR)

PR #1610, opened 2020-04-07
FUSETOOLS-3327 - upgrade internal wsdl2rest version

apupier:
Signed-off-by: Aurélien Pupier <apupier@redhat.com>

note versiosn of other dependencies can be found in this pom https://maven.repository.redhat.com/ga/org/jboss/fuse/wsdl2rest/wsdl2rest/0.8.0.fuse-760026-redhat-00001/wsdl2rest-0.8.0.fuse-760026-redhat-00001.pom

PR #1612, opened 2020-04-15
Revert "FUSETOOLS-3327 - upgrade internal wsdl2rest version"

apupier:
Reverts jbosstools/jbosstools-fuse#1610

it worked several times locally and on CI but is currently failing on master.

given the amount of time we have before the code freeze (which is today)
I think I will revert and we will ship without latest wsdl2rest
it was blocked on 7.3 version before anyway and nobody complained


apupier:
cannot use Github "revert" feature due to DCO check

PR #1613, opened 2020-04-15
Fusetools 3327 revert

apupier:
several builds locally and on CI on success... but it has just failed on master branch https://fusesource-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/jbosstools-fuse_master/1072/console

given the amount of time we have before the code freeze (which is today)
I think I will revert and we will ship without latest wsdl2rest
it was blocked on 7.3 version before anyway and nobody complained


apupier:
currently in code freeze so no merge possible.

When code unfrozen, would worth relaunching master. As repository.jboss.org [should have been fixed](https://redhat.service-now.com/surl.do?n=PNT0814030), this revert should be useless. If it is not fixed correctly, this revert still makes sense.

apupier:
finally no need of the revert.

we used the previous build for 11.6.0 release.
The problem was coming from a Maven repository outside of our codebase. It is now fixed.

PR #1630, opened 2020-09-08
FUSETOOLS-3368 - hide non-springboot projects with Camel 2.23.2

apupier:
- adapted UI test
- noticed that the list in SupportedCamelVersion is outdated. I won't
touch it in this PR to seprate concerns


djelinek (dismissed):

apupier:
status has not been updated but the PR jobs are green!!

The trick was that there was missing the bundle used as cache for Camel 2.23.2 for Fuse 7.7 (but again why it has not failed before???)

lhein (approved):

djelinek (approved):

bfitzpat (approved):

PR #1637, opened 2020-12-15
Avoid empty system editor to open in tests

apupier:
- had to split the test to have 2 differents projects (FuseProject rule)
used
- added a small additional check on the content of the file
- added to close the WizardDialog at the end of the test to have cleaner
test

The Java editor doesn't seem to be opened in the test. Not sure why.


lhein (approved):

lhein:
great, thanks for the fix

PR #1644, opened 2021-01-20
FUSETOOLS-3394 - upgrade grrovy lib to support JDK 14+

apupier:
the groovy lib doesn't match the camel version but this version has been
provided on Camel 2.25.x branch and it is working well so good chance
that it is not causing any problems.
JRE 15 is now shipped by default in Eclipse


bfitzpat (approved):
+1 pending green build

PR #1651, opened 2021-02-04
FUSETOOLS2-977 - use conditional wait for checking Problems view

apupier:
given that the computation of problems is done asynchronously through
Eclipse Jobs, it can be the cause of detected flakiness.


lhein (approved):

djelinek (approved):
Looks nice. Thank you Aurelien!

bfitzpat (approved):

PR #1658, opened 2021-02-09
FUSETOOLS2-980 - put hs_err_pid files generated in a place that can be

apupier:
archived

previously, it was generated inside the project under test. This project
is deleted at the end of the test and so we do not have access to the
hs_err_pid. Moving it a bit upper in the folder hierarchy will keep it.

@djelinek i'm not a fan of this solution but the easiest to implement. it would be nice if we can have a more common place to put that code. Or if we can provide parameters to tell where to generate the file directly. Any ideas?

### DIFF
--- a/uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/projectexplorer/CamelProject.java
+++ b/uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/projectexplorer/CamelProject.java
@@ -15,6 +15,7 @@ import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


lhein (approved):

djelinek (commented):
I see that the solution is not optimal but it looks like it costs minimal and if this is working as expected I would say go for it because it is prepared to use now.

More common place where to put code I see two options in uitests package
- move method into [FileUtils class](https://github.com/jbosstools/jbosstools-fuse/blob/master/uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/FileUtils.java)
- OR create new e.g. "Backup" class inside [utils](https://github.com/jbosstools/jbosstools-fuse/tree/master/uitests/plugins/org.jboss.tools.fuse.reddeer/src/org/jboss/tools/fuse/reddeer/utils)

I can investigate possibilities of backup custom files via RedDeer or implement more complex solution via parameters/requirements into Ui tests but it will cost more time and if it is not blocker no more for you I will let it open for now as improvement.

djelinek (approved):

apupier:
thanks @djelinek , I opened this task https://issues.redhat.com/browse/FUSETOOLS-3417 to improve it in a second iteration

PR #1663, opened 2021-02-10
FUSETOOLS-3416 - close editor to avoid stacktrace in log

apupier:
in practice, the editor is closed automatically but it is done in an
asynchronous way. In this test, it is complicated to wait for it and it
is not the purpose of this test. Consequently, a simple workaround is to
close the editor that the test has opened.


bfitzpat (approved):
+1 on green build. :)

lhein (approved):

PR #1664, opened 2021-02-10
FUSETOOLS2-991 - take screenshot on failure

apupier:
on Ci, it seems that the window sometimes lost focus. There should be
another application that is opened, taking the screenshot should allow
to determine what is it and help continuing investigation


bfitzpat (approved):
+1 makes sense

lhein (approved):

lhein:
I wouldn't be surprised if its the error popup regarding not being able to find a strict matching JRE

apupier:
> I wouldn't be surprised if its the error popup regarding not being able to find a strict matching JRE

I think it is the whole application that is losing focus because it is the `PlatformUI.getWorkbench().getDisplay().getActiveShell()` which is returnign null. If it is the popup for not strictly matching JRE, this popup would be the active shell.

PR #1670, opened 2021-02-11
FUSETOOLS2-991 - use first available shell if no active available

apupier:
randomly on CI, there is sometimes no active shell.


bfitzpat (approved):
Seems super weird that there wouldn't just be a default shell in the display instance. But I wonder what kind of trouble we're going to get by just picking the first available shell... hopefully none?

apupier:
> Seems super weird that there wouldn't just be a default shell in the display instance.

there are shells. But no active one.

> But I wonder what kind of trouble we're going to get by just picking the first available shell... hopefully none?

I do not think that the kind of shell is really important for this test. So I really hope there won't be any troubles.

lhein (approved):

PR #1671, opened 2021-03-02
FUSETOOLS-3437 - stop logging grape/Ivy during test execution

apupier:


bfitzpat (approved):
Less logging? Is that a good idea?

apupier:
> Less logging? Is that a good idea?

not very good but the best idea I have to migrate to new PSI jenkins and reuse Studio job configuration https://issues.redhat.com/browse/FUSETOOLS-3437?focusedCommentId=15879830&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15879830

lhein (approved):
given that we did stop seeing Ivy / Grape related issues I think this is ok.

PR #1674, opened 2021-03-04
FUSETOOLS-3446 - workaround to external issue of missing xsds

apupier:
see https://github.com/EclipseFdn/osgi.org/issues/11

### DIFF
--- a/core/tests/org.fusesource.ide.camel.tests.util/src/org/fusesource/ide/camel/tests/util/AbstractProjectCreatorRunnableIT.java
+++ b/core/tests/org.fusesource.ide.camel.tests.util/src/org/fusesource/ide/camel/tests/util/AbstractProjectCreatorRunnableIT.java
@@ -202,7 +202,11 @@ public abstract class AbstractProjectCreatorRunnableIT {
 		return marker -> {
 			try {
 				Object severity = marker.getAttribute(IMarker.SEVERITY);
-				return severity == null || severity.equals(IMarker.SEVERITY_ERROR);
+				boolean isError = severity == null || severity.equals(IMarker.SEVERITY_ERROR);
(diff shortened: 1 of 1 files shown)


bfitzpat (approved):
Should we put in a TODO or some comment to remind us to revert this at some point when we have a better fix?

apupier:
> Should we put in a TODO or some comment to remind us to revert this at some point when we have a better fix?

absolutely https://issues.redhat.com/browse/FUSETOOLS-3447 :-)

bjhargrave:
I have added the redirect from www.osgi.org to the xsd home on docs.osgi.org. So you probably don't need this work around any more.

apupier:
> I have added the redirect from [www.osgi.org](http://www.osgi.org) to the xsd home on docs.osgi.org. So you probably don't need this work around any more.

unfortunately, this is not working. it seems that Eclipse Lemminx is unable to follow the redirection.

apupier:
> > I have added the redirect from [www.osgi.org](http://www.osgi.org) to the xsd home on docs.osgi.org. So you probably don't need this work around any more.
> 
> unfortunately, this is not working. it seems that Eclipse Lemminx is unable to follow the redirection.

rectification. it has failed one time in 10 tries (but the first one)

apupier:
> > > I have added the redirect from [www.osgi.org](http://www.osgi.org) to the xsd home on docs.osgi.org. So you probably don't need this work around any more.
> > 
> > 
> > unfortunately, this is not working. it seems that Eclipse Lemminx is unable to follow the redirection.
> 
> rectification. it has failed one time in 10 tries (but the first one) locally.

I cannot reproduce anymore and it wasn't reporduced on CI so I guess I di d a bad manipulation initially.

PR #1675, opened 2021-03-04
FUSETOOLS-3348 - provide Bundle-Vendor when missing

apupier:
using i18n when already set up; otherwise put value directly

it avois 28 warnings in consloe log parsed to better see the other problems (and it is cleaner in general)


bfitzpat (approved):
I am very surprised that our bundle name/vendor settings were so spotty across all of these.

PR #1680, opened 2021-03-12
FUSETOOLS-3459 - avoid to call debugger actions when targeted process is terminated

apupier:


apupier:
there is unfortunately still the exception in the log.
I think that it remains interesting to add this code so that it can mitigate this desribe use case in title

bfitzpat:
If it doesn't work the way you were hoping, why do we want to merge the code?

apupier:
> If it doesn't work the way you were hoping, why do we want to merge the code?

I think that it is mitigating the case that the debugger tried to be disconnected although the process has already been terminated. It will fail faster as it won't try to to connect through JMX to a no more existing process.

bfitzpat:
If there is no noticeable change in behavior, how can you say whether that's actually the case or not? Unless the code change offers some appreciable, trackable metric (i.e. speed of build or size of log or change in actual behavior during runtime) I don't see the benefit of adding the code.

apupier:
> how can you say whether that's actually the case or not?

no proof, just understanding of the algorithm.

worked on that as it was potentially causing build failures on new PSI jenkins but it turns out it isn't the case.
Writing tests for this would require a bunch of hours.
Closing to focus on more critical work.

PR #1686, opened 2021-04-07
Fusetools 3465 pr job on GitHub action template

apupier:


.github/workflows/build_test-PR-templates.yml
```diff
@@ -0,0 +1,62 @@
+name: Build and Test Template tests on Pull Requests
```
bfitzpat, line 1: This seems like an odd name for the workflow. Are we testing the tests? Or testing the PR?

.github/workflows/publish_reports.yml
```diff
@@ -2,7 +2,7 @@ name: Publish reports
 
 on:
   workflow_run:
-    workflows: ["Build And Test"]
+    workflows: ["Build and Test main tests on Pull Requests", "Build and Test Template tests on Pull Requests"]
```
bfitzpat, line 5: This seems like an odd name for the workflow. Are we testing the tests? Or testing the PR?

.github/workflows/build_test-PR-templates.yml
apupier: we are testing the product when a pull request is created

.github/workflows/publish_reports.yml
apupier: we are testing the product when a pull request is created

bfitzpat (approved):

PR #1695, opened 2021-06-14
Fusetools 3481 use specific camel xsd versions

apupier:
use specific xsd versions:
- 2.17.7 xsd for Fuse 6 templates
- 2.21.5 for Fuse 7.0+ and 7.6-
- 2.23.3 for Fuse 7.7+

it avoids to have "latest" to be picked that corresponds to Camel 3.x and which no whave incompatible changes


bfitzpat (approved):
I always wished there was a better way to do this than specifying the version in the schema url, but hey I can't change the world. :) Seems to build green and be happier

lhein (approved):

PR #1712, opened 2021-09-06
FUSETOOLS-3504 - use registry.redhat.io for Fuse 7.9

apupier:
given that anonymous access seems to have been moved from
registry.access.redhat.com to registry.redhat.io


djelinek:
Hi @apupier I found one more occurrence here - https://github.com/jbosstools/jbosstools-fuse/blob/843f157c586c07cc2ddcf99561612c83a59010cd/editor/plugins/org.fusesource.ide.projecttemplates/template-sources/simple/ose/simple-ose-log-springboot-fuse7.9/pom.xml#L94 , do we need modify also this url to be safe?

apupier:
> do we need modify also this url to be safe?

I do not know but we need to check. Seems high probability that we should modify it too

djelinek (approved):

lhein (approved):

PR #1749, opened 2021-12-23
Fusetools 3545 add to wizard

apupier:


apupier:
merging without review as he is on PTo and it will allow to have Fuse 7.10 support in January instead of April (or even later)

PR #1761, opened 2022-03-01
FUSETOOLS-3611 - Upgrade wsdl2rest internal lib to fuse 7.10.1

apupier:


apupier:
the wsdl2rest feature will be removed https://github.com/jbosstools/jbosstools-fuse/pull/1762 so no need to upgrade

PR #1766, opened 2022-03-02
Convert projects to Maven projects

apupier:
it will ease import of projects in a new workspace


apupier:
project is no more maintained for quite some time now

PR #1775, opened 2022-05-17
FUSETOOLS-3649 - remove org.jboss.tools.common.jaxb

apupier:
the idea is to use built-in jaxb functionalities which should be now
compatible with Java 11+. Also we do not have to be compatible with Java
8 anymore


apupier:
several failures with error:
```
java.lang.NoClassDefFoundError: javax/xml/bind/annotation/XmlElementDecl
	at org.jboss.tools.fuse.transformation.core.model.xml.XmlModelGeneratorIT.generateFromInstanceWithMultipleNamespaces(XmlModelGeneratorIT.java:91)
Caused by: java.lang.ClassNotFoundException: javax.xml.bind.annotation.XmlElementDecl cannot be found by org.jboss.tools.fuse.transformation.core_11.15.0.v20220517-0952
	at org.jboss.tools.fuse.transformation.core.model.xml.XmlModelGeneratorIT.generateFromInstanceWithMultipleNamespaces(XmlModelGeneratorIT.java:91)
```

transformation/plugins/org.jboss.tools.fuse.transformation.core/META-INF/MANIFEST.MF
```diff
@@ -20,8 +20,12 @@ Require-Bundle: org.eclipse.core.runtime;bundle-version="3.11.0",
  org.fusesource.ide.foundation.core;bundle-version="10.0.0",
  org.fusesource.ide.foundation.ui;bundle-version="10.0.0",
  org.fusesource.ide.camel.editor;bundle-version="10.0.0",
- org.jboss.tools.common.jaxb;resolution:=optional,
  org.fusesource.ide.jsonschema2pojo-0.4.13;bundle-version="11.10.0"
+Import-Package: com.sun.codemodel,
+ com.sun.tools.xjc.api,
```
apupier, line 25: this is currently provided from embedded libs jaxb-xjc from org.jboss.tools.common.jaxb.

Not sure what is replacing it, seems the jakarta.xml.bind is not providing it.
There is also an old org.jboss.tool.locus.jaxb-xjc but I think we do not want to have locus anymore.

So which other bundle?

@sbouchet any ideas?

sbouchet: you're correct, both packages were provided by common.jaxb, coming from one single jar : 
jaxb-xjc-2.2.6.jar

as it's not part of any current p2 repo, i just suggest to introduce it back to this bundle directly as it seems it's the only usage ? ( and export it for your other bundle that need it ) 
it should be easy as copy/paste from common.jaxb :
						<artifactItem>
							<groupId>com.sun.xml.bind</groupId>
							<artifactId>jaxb-xjc</artifactId>
							<version>2.2.6</version>
						</artifactItem>

add it to the existing copy-deps and you should be good to go.

also, when maven repo will be available in JBossTools TP, it will be simple to get from central: 
https://mvnrepository.com/artifact/com.sun.xml.bind/jaxb-xjc

apupier: - it is used in 2 other bundles jsonschemapojo and o.j.t.f.transformation.core.editor
- Requires to upgrade to jaxb-xjc 2.3.3 as the bundle jakarta.xml.bind is 2.3.3 and is incompatible with 2.2.6
- Needed to add a lot more of import-packages

see https://github.com/jbosstools/jbosstools-fuse/pull/1775/commits/c3f1cf1667989565ded46605f21977f58972b4d1

djelinek (approved):

sbouchet (approved):
Well done !

PR #1792, opened 2022-10-11
FUSETOOLS2-3665 - fix deployment to Karaf

apupier:
- changed only for latest versions of examples to limit things to be tested. Maybe we can live with the regression for older versions as there are workarounds and the JBoss Tools is not part of the support. users can stick with older version of CodeReady Studio for older Fuse version?
- the maven-bundle-plugin on Camel fuse supported version is 3.5.1 https://github.com/jboss-fuse/camel/blob/f96b861d57b660710011110c991a37e272c1186d/parent/pom.xml#L488 but only 5.x contains a fix. So we have not aligned version. In theory, it is affecting only the build and not the runtime so should be fine.
- The m2e connector which was working before was in version 0.8.1 (sonatype.tycho), it was overriding the maven-bundle-plugin embedded m2e connector configuration. Unfortunately, this sonatype.tycho has been abandoned and is not working with latest Eclipse 2022-09 (incompatible with M2e 2.x)
- note that an extra configuration must be provided to have incremental build generating the Manifest.mf correctly.


apupier:
this PR will fix the problem for newly created project. For exiting one, still need to modify the pom.

joshiraez (approved):

PR #1795, opened 2022-10-27
FUSETOOLS-3666 - filter DataFormats Based on Runtime Specific Catalog In

apupier:
Details Section

given that the EIPs are providing potential dataformats without care of the runtime specific context


apupier:
on hold awaiting answer to know if it is a correct solution at functional level, I.e. that the dataformats must be filtered https://chat.google.com/room/AAAA57qndy4/pJLr2nMMxe8

apupier:
these dataformats are in fact supported and should nto be filtered out. Reported https://issues.redhat.com/browse/ENTESB-20175

